PR #1, opened 2021-07-21
nits on ERC20.sol

Ro5s:
lil space nit 😛// also we can save some gas by doing 'unchecked' for resulting balances from `transfer` ~ `transferFrom`? (since we checked sender balance with implicit safeMath)

### DIFF
--- a/src/erc20/ERC20.sol
+++ b/src/erc20/ERC20.sol
@@ -73,6 +73,7 @@ contract ERC20 {
     /*///////////////////////////////////////////////////////////////
                               ERC20 LOGIC
     //////////////////////////////////////////////////////////////*/
+    
     function approve(address spender, uint256 value) external returns (bool) {
         allowance[msg.sender][spender] = value;
 
@@ -83,7 +84,7 @@ contract ERC20 {
 
     function transfer(address to, uint256 value) external returns (bool) {
         balanceOf[msg.sender] -= value;
-        balanceOf[to] += value;
+        unchecked { balanceOf[to] += value; }
 
         emit Transfer(msg.sender, to, value);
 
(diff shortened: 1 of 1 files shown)


TransmissionsDev:
> lil space nit 😛// also we can save some gas by doing 'unchecked' for resulting balances from `transfer` ~ `transferFrom`? (since we checked sender balance with implicit safeMath)

ayy awesome to have a gigachad like u checkin out the lib (still very WIP btw 😓)

maybe it's just cuz i'm fried af but this is't necessarily fully safe is it? if the `from` user has `(type(uint256).max / 2) + 1` as their balance and the `to` user has `(ype(uint256).max / 2` as their balance, the `to` user's balance would overflow but the transaction would not revert (as it would not cause an underflow to set the `from` user's balance back down to 0?)

Ro5s:
hey hey ~~ no chads, just simple code here ;0 // but that's a good check.... also bit fried and will take another look.

but, wondering if we can avoid the `from` user from having a balance that could potentially overflow by containing everything on `mint` .... as in, we wouldn't be able to have users who could overflow each other?

TransmissionsDev:
> hey hey ~~ no chads, just simple code here ;0 // but that's a good check.... also bit fried and will take another look.
> 
> but, wondering if we can avoid the `from` user from having a balance that could potentially overflow `((type(uint256).max / 2) + 1) ` by containing things with safeMath in `mint` (the total is safe.... and users shouldn't be able to swap among each other to overflow?)

yeah interesting we could cap totalSupply at `type(uint256).max / 2 `?

TransmissionsDev:
ahhh shit yeah ur right! 

totalSupply being increased with checked math on mint will prevent user balance sums from going above `type(uint256).max`, which makes the scenario i described above impossible!

TransmissionsDev:
thx so much @Ro5s :)

PR #2, opened 2021-08-08
fix: Approval emit missing 'owner' in ERC20.sol

Ro5s:
teensy tiny nits on `Approval` emit :~)


TransmissionsDev:
Damn how did that make it in :/

Probably would have caught earlier if dapptools had event emission testing. 

Thanks for the fix 🙏

PR #6, opened 2021-09-06
✨ ERC721

Ro5s:
add minimalist ERC-721 token for consideration with Solmate stylings ✔️

### DIFF
--- a/src/erc721/ERC721.sol
+++ b/src/erc721/ERC721.sol
@@ -1,14 +1,13 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 pragma solidity >=0.8.0;
 
-/// @notice Modern and gas efficient ERC-721 + EIP-2612-like implementation.
-// ! TO DO - add safeTransfer stuff 🙏
+/// @notice Modern and gas efficient ERC-721 + ERC-20/EIP-2612-like implementation.
 contract ERC721 {
     /*///////////////////////////////////////////////////////////////
                                   EVENTS
     //////////////////////////////////////////////////////////////*/
     
-    event Transfer(address indexed from, address indexed to, uint256 indexed tokenId);
+    event Transfer(address indexed owner, address indexed to, uint256 indexed tokenId);
     
     event Approval(address indexed owner, address indexed spender, uint256 indexed tokenId);
     
@@ -39,11 +38,11 @@ contract ERC721 {
     mapping(address => mapping(address => bool)) public isApprovedForAll;
     
     /*///////////////////////////////////////////////////////////////
-                         PERMIT/EIP-2612 STORAGE
+                         PERMIT/EIP-2612-LIKE STORAGE
     //////////////////////////////////////////////////////////////*/
     
     bytes32 public constant PERMIT_TYPEHASH =
-        keccak256("Permit(address owner,address spender,uint256 tokenId,uint256 nonce,uint256 deadline)");
+        keccak256("Permit(address spender,uint256 tokenId,uint256 nonce,uint256 deadline)");
         
     bytes32 public immutable DOMAIN_SEPARATOR;
 
@@ -68,21 +67,44 @@ contract ERC721 {
     }
     
     /*///////////////////////////////////////////////////////////////
+                              ERC-20-LIKE LOGIC
+    //////////////////////////////////////////////////////////////*/
+    
+    function transfer(address to, uint256 tokenId) external payable {
+        require(msg.sender == ownerOf[tokenId], "NOT_OWNER");
+        
+        // This is safe because ownership is checked
+        // against decrement, and sum of all user
+        // balances can't exceed type(uint256).max!
+        unchecked {
+            balanceOf[msg.sender]--; 
+        
+            balanceOf[to]++;
+        }
+        
+        delete getApproved[tokenId];
+        
+        ownerOf[tokenId] = to;
+        
+        emit Transfer(msg.sender, to, tokenId); 
+    }
+    
+    /*///////////////////////////////////////////////////////////////
                               ERC-721 LOGIC
     //////////////////////////////////////////////////////////////*/
     
-    function supportsInterface(bytes4 sig) external pure returns (bool) {
-        return (sig == 0x80ac58cd || sig == 0x5b5e139f);
+    function supportsInterface(bytes4 interfaceId) external pure returns (bool supported) {
+        supported = interfaceId == 0x80ac58cd || interfaceId == 0x5b5e139f;
     }
     
-    function approve(address spender, uint256 tokenId) external {
+    function approve(address spender, uint256 tokenId) external payable {
         address owner = ownerOf[tokenId];
         
-        require(msg.sender == owner || isApprovedForAll[owner][msg.sender], "NOT_AUTHORIZED");
+        require(msg.sender == owner || isApprovedForAll[owner][msg.sender], "NOT_APPROVED");
         
         getApproved[tokenId] = spender;
         
-        emit Approval(msg.sender, spender, tokenId); 
+        emit Approval(owner, spender, tokenId); 
     }
     
     function setApprovalForAll(address operator, bool approved) external {
@@ -91,45 +113,55 @@ contract ERC721 {
         emit ApprovalForAll(msg.sender, operator, approved);
     }
 
-    function transfer(address to, uint256 tokenId) external {
-        require(msg.sender == ownerOf[tokenId], "NOT_OWNER");
-        
-        balanceOf[msg.sender]--; 
-        
-        balanceOf[to]++; 
-        
-        getApproved[tokenId] = address(0);
-        
-        ownerOf[tokenId] = to;
-        
-        emit Transfer(msg.sender, to, tokenId); 
-    }
-    
-    function transferFrom(address, address to, uint256 tokenId) external {
+    function transferFrom(address, address to, uint256 tokenId) public payable {
         address owner = ownerOf[tokenId];
         
         require(
-            msg.sender == owner || getApproved[tokenId] == msg.sender 
-            || isApprovedForAll[owner][msg.sender], "NOT_AUTHORIZED"
+            msg.sender == owner 
+            || msg.sender == getApproved[tokenId]
+            || isApprovedForAll[owner][msg.sender], 
+            "NOT_APPROVED"
         );
         
-        balanceOf[owner]--; 
+        // This is safe because ownership is checked
+        // against decrement, and sum of all user
+        // balances can't exceed type(uint256).max!
+        unchecked { 
+            balanceOf[owner]--; 
         
-        balanceOf[to]++;
+            balanceOf[to]++;
+        }
         
-        getApproved[tokenId] = address(0);
+        delete getApproved[tokenId];
         
         ownerOf[tokenId] = to;
         
         emit Transfer(owner, to, tokenId); 
     }
     
+    function safetransferFrom(address, address to, uint256 tokenId) external payable {
+        safetransferFrom(address(0), to, tokenId, "");
+    }
+    
+    function safetransferFrom(address, address to, uint256 tokenId, bytes memory data) public payable {
+        transferFrom(address(0), to, tokenId); 
+        
+        if (to.code.length != 0) {
+            // selector = `onERC721Received(address,address,uint,bytes)`
+            (, bytes memory returned) = to.staticcall(abi.encodeWithSelector(0x150b7a02,
+                msg.sender, address(0), tokenId, data));
+                
+            bytes4 selector = abi.decode(returned, (bytes4));
+            
+            require(selector == 0x150b7a02, "NOT_ERC721_RECEIVER");
+        }
+    }
+    
     /*///////////////////////////////////////////////////////////////
-                          PERMIT/EIP-2612 LOGIC
+                          PERMIT/EIP-2612-LIKE LOGIC
     //////////////////////////////////////////////////////////////*/
 
     function permit(
-        address owner,
         address spender,
         uint256 tokenId,
         uint256 deadline,
@@ -138,22 +170,24 @@ contract ERC721 {
         bytes32 s
     ) external {
         require(deadline >= block.timestamp, "PERMIT_DEADLINE_EXPIRED");
-        // This is reasonably safe because incrementing past type(uint256).max
-        // through approvals is exceedingly unlikely!
-        unchecked {
-            bytes32 digest = keccak256(
-                abi.encodePacked(
-                    "\x19\x01",
-                    DOMAIN_SEPARATOR,
-                    keccak256(abi.encode(PERMIT_TYPEHASH, owner, spender, tokenId, nonces[owner]++, deadline))
-                )
-            );
+        
+        address owner = ownerOf[tokenId];
+        
+        bytes32 digest = keccak256(
+            abi.encodePacked(
+                "\x19\x01",
+                DOMAIN_SEPARATOR,
+                keccak256(abi.encode(PERMIT_TYPEHASH, spender, tokenId, nonces[owner]++, deadline))
+            )
+        );
 
-            address recoveredAddress = ecrecover(digest, v, r, s);
-            require(recoveredAddress != address(0) && recoveredAddress == owner, "INVALID_PERMIT_SIGNATURE");
-            require(recoveredAddress == ownerOf[tokenId], "INVALID_OWNER");
-        }
-
+        address recoveredAddress = ecrecover(digest, v, r, s);
+        require(recoveredAddress != address(0) 
+                && recoveredAddress == owner
+                || isApprovedForAll[owner][recoveredAddress], 
+                "INVALID_PERMIT_SIGNATURE"
+        );
+        
         getApproved[tokenId] = spender;
 
         emit Approval(owner, spender, tokenId);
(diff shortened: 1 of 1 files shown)


src/erc721/ERC721.sol
```diff
@@ -0,0 +1,193 @@
...
+        uint256 tokenId,
+        uint256 deadline,
+        uint8 v,
+        bytes32 r,
+        bytes32 s
+    ) external {
+        require(deadline >= block.timestamp, "PERMIT_DEADLINE_EXPIRED");
+        // This is reasonably safe because incrementing past type(uint256).max
+        // through approvals is exceedingly unlikely!
+        unchecked {
```
transmissions11, line 143: can we limit the unchecked to just where the nonce is incremented (idk if u can wrap just around nonces++ but at least only the digest block

transmissions11: also should do this with ERC20 if we're doing here

src/erc721/ERC721.sol
```diff
@@ -0,0 +1,193 @@
...
+        address owner,
+        address spender,
+        uint256 tokenId,
+        uint256 deadline,
+        uint8 v,
+        bytes32 r,
+        bytes32 s
+    ) external {
+        require(deadline >= block.timestamp, "PERMIT_DEADLINE_EXPIRED");
+        // This is reasonably safe because incrementing past type(uint256).max
```
transmissions11, line 141: nit: space above

src/erc721/ERC721.sol
```diff
@@ -0,0 +1,193 @@
...
+
+        emit Approval(owner, spender, tokenId);
+    }
+    
+    /*///////////////////////////////////////////////////////////////
+                          INTERNAL UTILS
+    //////////////////////////////////////////////////////////////*/
+    
+    function mint(address to, uint256 tokenId, string calldata _tokenURI) internal { 
+        require(ownerOf[tokenId] == address(0), "TOKEN_ID_NOT_UNIQUE");
```
transmissions11, line 167: mebbe just `ALREADY_MINTED`?

src/erc721/ERC721.sol
```diff
@@ -0,0 +1,193 @@
...
+        
+        totalSupply++;
+        
+        balanceOf[to]++;
```
transmissions11, line 171: can this be unchecked?

src/erc721/ERC721.sol
```diff
@@ -0,0 +1,193 @@
...
+        
+        tokenURI[tokenId] = _tokenURI;
+        
+        emit Transfer(address(0), to, tokenId); 
+    }
+    
+    function burn(address from, uint256 tokenId) internal { 
+        require(ownerOf[tokenId] == from, "NOT_FROM_OWNER");
+        
+        totalSupply--;
```
transmissions11, line 183: unchecked?

src/erc721/ERC721.sol
```diff
@@ -0,0 +1,193 @@
...
+        
+        ownerOf[tokenId] = to;
+        
+        tokenURI[tokenId] = _tokenURI;
+        
+        emit Transfer(address(0), to, tokenId); 
+    }
+    
+    function burn(address from, uint256 tokenId) internal { 
+        require(ownerOf[tokenId] == from, "NOT_FROM_OWNER");
```
transmissions11, line 181: whats the point of accepting `from` here at all then if we do this storage load?

src/erc721/ERC721.sol
```diff
@@ -0,0 +1,193 @@
...
+        
+        totalSupply--;
+        
+        balanceOf[from]--;
+        
+        ownerOf[tokenId] = address(0);
+        
+        tokenURI[tokenId] = "";
```
transmissions11, line 189: mebbe use delete here for explict

src/erc721/ERC721.sol
```diff
@@ -0,0 +1,193 @@
...
+    }
+    
+    function burn(address from, uint256 tokenId) internal { 
+        require(ownerOf[tokenId] == from, "NOT_FROM_OWNER");
+        
+        totalSupply--;
+        
+        balanceOf[from]--;
+        
+        ownerOf[tokenId] = address(0);
```
transmissions11, line 187: mebbe use delete here for explict

src/erc721/ERC721.sol
```diff
@@ -0,0 +1,193 @@
...
+        address owner = ownerOf[tokenId];
+        
+        require(
+            msg.sender == owner || getApproved[tokenId] == msg.sender 
+            || isApprovedForAll[owner][msg.sender], "NOT_AUTHORIZED"
+        );
+        
+        balanceOf[owner]--; 
+        
+        balanceOf[to]++;
```
transmissions11, line 118: unchecked?

src/erc721/ERC721.sol
```diff
@@ -0,0 +1,193 @@
...
+        
+        getApproved[tokenId] = address(0);
```
transmissions11, line 120: can we use `delete` to be explicit?

src/erc721/ERC721.sol
```diff
@@ -0,0 +1,193 @@
...
+            )
+        );
+    }
+    
+    /*///////////////////////////////////////////////////////////////
+                              ERC-721 LOGIC
+    //////////////////////////////////////////////////////////////*/
+    
+    function supportsInterface(bytes4 sig) external pure returns (bool) {
+        return (sig == 0x80ac58cd || sig == 0x5b5e139f);
```
transmissions11, line 75: is this required?

src/erc721/ERC721.sol
```diff
@@ -0,0 +1,193 @@
...
+    
+    uint256 public totalSupply;
+    
+    mapping(address => uint256) public balanceOf;
+    
+    mapping(uint256 => address) public ownerOf;
+    
+    mapping(uint256 => string) public tokenURI;
+    
+    mapping(uint256 => address) public getApproved;
```
transmissions11, line 37: is this the standard name?

transmissions11:
wow this is awesome tysm!! left some nit requests

src/erc721/ERC721.sol
transmissions11: mehhh idk i'd rather just remove this unchecked. not worth it here imo— all the other instances we use it at are sound

Ro5s: yeah, so the `unchecked` bracketing seemingly scopes away the `digest` variable, so need to wrap underlying. But agree, might as well keep things totally safe -- will remove :~)

src/erc721/ERC721.sol
Ro5s: yeah, I guess this safety check doesn't do that much at base level -- can just remove the storage load, as well. The idea here though, is basically ... want predictability, so we don't have Alice as `from` param and end up burning a token Alice doesn't actually own. But, I can see how this can be opinionated, and controls can be put into inheritance.

src/erc721/ERC721.sol
Ro5s: yeah, though I find it tedious hah (like anyone can claim their contract supports an interface without actually implementing....) But to the point ~~ [ERC-721](https://eips.ethereum.org/EIPS/eip-721) supports [ERC-165](https://eips.ethereum.org/EIPS/eip-165) 'standard interface detection', 
image

Ro5s: here, we flag that ERC721 implementation supports `ERC721` and "metadata extension" 
image

ERC721 - `0x80ac58cd`
image

Metadata - `0x5b5e139f`
image

This implementation also partially supports "enumeration extension" (by ref to `totalSupply()`, but I have not flagged that in ERC-165 interface, as the indexing functions don't seem all that useful, and have not included,
image

src/erc721/ERC721.sol
```diff
@@ -1,14 +1,13 @@
...
-/// @notice Modern and gas efficient ERC-721 + EIP-2612-like implementation.
-// ! TO DO - add safeTransfer stuff 🙏
+/// @notice Modern and gas efficient ERC-721 + ERC-20/EIP-2612-like implementation.
 contract ERC721 {
     /*///////////////////////////////////////////////////////////////
                                   EVENTS
     //////////////////////////////////////////////////////////////*/
     
-    event Transfer(address indexed from, address indexed to, uint256 indexed tokenId);
+    event Transfer(address indexed owner, address indexed to, uint256 indexed tokenId);
```
transmissions11, line 10: hmm i like `from`

src/erc721/ERC721.sol
```diff
@@ -67,22 +66,45 @@ contract ERC721 {
...
     //////////////////////////////////////////////////////////////*/
     
-    function supportsInterface(bytes4 sig) external pure returns (bool) {
-        return (sig == 0x80ac58cd || sig == 0x5b5e139f);
+    function supportsInterface(bytes4 interfaceId) external pure returns (bool supported) {
+        supported = interfaceId == 0x80ac58cd || interfaceId == 0x5b5e139f;
     }
     
-    function approve(address spender, uint256 tokenId) external {
+    function approve(address spender, uint256 tokenId) external payable {
```
transmissions11, line 100: why payable?

Ro5s: tbh, I think `payable` was an odd choice for EIP authors. I believe they wanted to support royalty payments, and see 'deeds' or digital art as more likely to take advantage of 'fee-on-transfer' functions. Can punt this into inheritance, but just added to flag. 
image

It is also especially odd to include payability on approvals..... as the fee should be grabbed on `transferFrom`.....

src/erc721/ERC721.sol
Ro5s: yeah, I actually mulled this for an inordinate amount of time lol. But, yea 'from' pairs better to 'to' and would match up with ERC20 .... foine foine ;0

abcoathup (commented):
Default behavior should be a tokenURI generated from baseURI + tokenID, saving on the gas cost of storing a tokenURI per tokenID.  See: https://github.com/OpenZeppelin/openzeppelin-contracts/issues/1745

This allows for decentralized IPFS (directory) as baseURI and tokenID holding JSON, or centralized server and tokenID returning JSON.

See OpenZeppelin Contracts Wizard (https://wizard.openzeppelin.com/#erc721)
```solidity
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.2;

import "@openzeppelin/contracts/token/ERC721/ERC721.sol";
import "@openzeppelin/contracts/access/Ownable.sol";
import "@openzeppelin/contracts/utils/Counters.sol";

contract MyToken is ERC721, Ownable {
    using Counters for Counters.Counter;

    Counters.Counter private _tokenIdCounter;

    constructor() ERC721("MyToken", "MTK") {}

    function _baseURI() internal pure override returns (string memory) {
        return "https://example.com/tokens/";
    }

    function safeMint(address to) public onlyOwner {
        _safeMint(to, _tokenIdCounter.current());
        _tokenIdCounter.increment();
    }
}
```

tokenURI function can be overridden so that JSON and SVG can be provided purely on chain.

abcoathup:
Depending if the contract was inspired by another implementation, you could add a reference to the actual contracts and what license they use.

abcoathup:
I suggest adding a link to this repo to the comments of the ERC721.sol so they don't get lost if the Solidity is copy/pasted

src/erc721/ERC721.sol
```diff
@@ -0,0 +1,237 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity >=0.8.0;
+
+/// @notice Modern and gas efficient ERC-721 + ERC-20/EIP-2612-like implementation.
+contract ERC721 {
+    /*///////////////////////////////////////////////////////////////
```
platocrat, line 6: these documentation lines 🤢

Ro5s: do you mean these guys?
`/*///////////////////////////////////////////////////////////////` 

I blame transmissions :~0 But have grown on me

transmissions11: I overuse them a tad in solmate might strip em out except for the more complex contracts 😅

src/erc721/ERC721.sol
dievardump: I think you are also supposed to add the ERC165 interface too actually, to be totally compliant with 165

```solidity
contract ERC165 is IERC165 {
    /*
     * bytes4(keccak256('supportsInterface(bytes4)')) == 0x01ffc9a7
     */
    bytes4 private constant _INTERFACE_ID_ERC165 = 0x01ffc9a7;
```

src/erc721/ERC721.sol
```diff
@@ -0,0 +1,237 @@
...
+        ownerOf[tokenId] = to;
+        
+        emit Transfer(owner, to, tokenId); 
+    }
+    
+    function safetransferFrom(address, address to, uint256 tokenId) external {
+        safetransferFrom(address(0), to, tokenId, "");
+    }
+    
+    function safetransferFrom(address, address to, uint256 tokenId, bytes memory data) public {
```
dievardump, line 146: Both safeTransferFrom have a typo on the T which should be uppercase

src/erc721/ERC721.sol
```diff
@@ -0,0 +1,237 @@
...
+    ) external {
+        require(deadline >= block.timestamp, "PERMIT_DEADLINE_EXPIRED");
+        
+        address owner = ownerOf[tokenId];
+        
+        bytes32 digest = keccak256(
+            abi.encodePacked(
+                "\x19\x01",
+                DOMAIN_SEPARATOR,
+                keccak256(abi.encode(PERMIT_TYPEHASH, spender, tokenId, nonces[owner]++, deadline))
```
dievardump, line 180: For NFTs, the nonce should probably be more `mapping(uint256 => uint256)` because it's for a tokenId.

Your implementation will invalidate any permit on any other id of this owner, once a permit is used.

Whenever we should here invalidate any other permit on this tokenId, not from this owner.

src/erc721/ERC721.sol
transmissions11: Good catch. @Ro5s make sure to update TridentNFT too! :p

src/erc721/ERC721.sol
Ro5s: @dievardump nice check here. I don't think we need this level of introspection, as similar to [OZ erc721](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/ERC721.sol)(which imports their `ERC165` contract, but not extra meta, `ERC165Checker` library...), we implement the function required by ERC165 to confirm erc721 and extension support 
image. I am curious of cases where an app or other contract might call `supportsInterface`, the function contemplated by erc165, and also wants to make sure it supports this same interface on meta level?

src/erc721/ERC721.sol
Ro5s: lazer eyed @dievardump -- will make those changes :~0)

src/erc721/ERC721.sol
Ro5s: good catch @dievardump :~)!

Ro5s:
should we pre-hash this bad boi? 
image

platocrat:
> should we pre-hash this bad boi?
> image

yes? What's a pre-hash? Is that just assigning `PERMIT_TYPEHASH` to the value of keccak256 hash of `"Permit(address spender,uint256 tokenId,uint256 nonce,uint256 deadline)"` so that you don't have to call `keccak256()` ?

transmissions11:
> should we pre-hash this bad boi?

nah

platocrat:
what's a pre-hash?

transmissions11:
> Is that just assigning `PERMIT_TYPEHASH` to the value of keccak256 hash of `"Permit(address spender,uint256 tokenId,uint256 nonce,uint256 deadline)"` so that you don't have to call `keccak256()` ?

Yea but since the variable is constant the optimizer does that for us anyway. In old versions of solc I think it made a tiny difference to precompute ([see this PR](https://github.com/Uniswap/v2-core/pull/64)) but even then it wasn't worth imo— and i think the compiler has its shit worked out now

platocrat:
> > Is that just assigning `PERMIT_TYPEHASH` to the value of keccak256 hash of `"Permit(address spender,uint256 tokenId,uint256 nonce,uint256 deadline)"` so that you don't have to call `keccak256()` ?
> 
> Yea but since the variable is constant the optimizer does that for us anyway. In old versions of solc I think it made a tiny difference to precompute ([see this PR](https://github.com/Uniswap/v2-core/pull/64)) but even then it wasn't worth imo— and i think the compiler has its shit worked out now

I see. Thanks for explaining :peepolove:

platocrat:
is there anything missing before this can be merge?

transmissions11:
dont rly wanna merge until we have tests but dont have enough time to do myself rn

if anyone wants to write em i'd be happy to merge after

https://github.com/brianmcmichael/ds-deed/blob/master/src/deed.t.sol <-- can use this as a starting point but ideally we have some symbolic/invariant/fuzz tests too

platocrat:
@z0r0z working on this :)

will likely have it done by tn

platocrat:
@Ro5s i think you forgot to add an `isApprovedForAll` function to test that a user is approved

transmissions11:
this repo may be helpful: https://github.com/t4sk/erc721/blob/main/src/test/ERC721.t.sol

z0r0z:
this https://github.com/Rari-Capital/solmate/pull/63 kinda deprecates this

transmissions11:
coolio

PR #24, opened 2021-09-30
♻️ Convert external erc20 methods to public

jparklev:
Methods touched are: `approve`, `transfer`, `transferFrom`, and `permit`.

Motivations for this PR are that it:
* Conforms to the [ERC20 standard](https://eips.ethereum.org/EIPS/eip-20) 
* Aligns with [OpenZeppelin's ERC20 implementation](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/ERC20.sol)
* Allows inheriting contracts to call these methods with `super`, enabling more flexible usage (e.g. for aave-style tokens that add an extra call on every `transfer`)

A question is if there are extra gas costs to using `public` here, but my understanding is that the arguments ([1](https://gus-tavo-guim.medium.com/public-vs-external-functions-in-solidity-b46bcf0ba3ac), [2](https://ethereum.stackexchange.com/questions/19380/external-vs-public-best-practices)) for that only apply to reference-type arguments.

### DIFF
--- a/src/erc20/ERC20.sol
+++ b/src/erc20/ERC20.sol
@@ -67,7 +67,7 @@ contract ERC20 {
                               ERC20 LOGIC
     //////////////////////////////////////////////////////////////*/
 
-    function approve(address spender, uint256 value) external returns (bool) {
+    function approve(address spender, uint256 value) public returns (bool) {
         allowance[msg.sender][spender] = value;
 
         emit Approval(msg.sender, spender, value);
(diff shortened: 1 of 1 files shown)


transmissions11:
Down to merge just need to:

- Verify the gas snapshots don’t change (I trust u hehe just need to be 100% sure)
- Add some dev comments explaining that although the functions are public they’re not used internally— only provided for child contracts 

Also do you want to mark these virtual too?

jparklev:
> Also do you want to mark these virtual?

Yep, great shout – brb

Issue #10, opened 2021-09-18
🔒 Optimize validating chainID in `permit` 


transmissions11:
[<img width="679" alt="Screen Shot 2021-09-18 at 1 01 19 AM" src="https://user-images.githubusercontent.com/26209401/133881468-1efc743c-6b30-46c0-9e76-9872e6573298.png">](https://github.com/trailofbits/publications/blob/master/reviews/YieldProtocol.pdf)

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/draft-EIP712.sol
https://github.com/yieldprotocol/yield-utils-v2/blob/main/contracts/token/ERC20Permit.sol
https://github.com/boringcrypto/BoringSolidity/blob/master/contracts/Domain.sol

transmissions11:
Using the whole caching method in the examples above adds a lot of code, wonder how much gas that saves or if we just recompute each time like in DappsysV2: 

https://github.com/dapp-org/dappsys-v2/blob/a59e20576ef9febcd8ac350f8fad363e55c5277a/src/token.sol#L96-L98

transmissions11:
Links to read up on later: 

https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md

https://docs.soliditylang.org/en/v0.6.2/abi-spec.html#abi

transmissions11:
Is this whole hashStruct thing they've got in 712 how structs are actually encoded?

also the 712 doc has a more optimized hashStruct impl we should check out

PR #27, opened 2021-10-01
add replay protection to ERC20

Ro5s:
for consideration ~~ add `_calculateDomainSeparator()` and related variables to protect against signature replay across forks

### DIFF
--- a/src/erc20/ERC20.sol
+++ b/src/erc20/ERC20.sol
@@ -39,7 +39,9 @@ contract ERC20 {
     bytes32 public constant PERMIT_TYPEHASH =
         keccak256("Permit(address owner,address spender,uint256 value,uint256 nonce,uint256 deadline)");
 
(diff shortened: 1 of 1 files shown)


transmissions11:
Closes https://github.com/Rari-Capital/solmate/issues/10

transmissions11:
For any external observers: Waiting to merge so we can think a bit more if there's a cleaner way to do this. 

Wanted something like this:

image

But you can't read an immutable before they're initialized.

Also worth evaluating whether we should go all out with the immutable caching to lower gas usage in the event of a fork, in exchange for increasing complexity:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/draft-EIP712.sol

I think the answer is it's not worth it but just putting it out there :p

transmissions11:
implemented directly in master sorry boo

PR #44, opened 2021-10-31
⚡️ Optimize Auth

transmissions11:


transmissions11:
Want to hold off on merging until we have a good set of fuzz tests for Auth and all the authorities that fuzz function sigs, callers, and targets. Should do this on isAuthorized and canCall

src/auth/Auth.sol
```diff
@@ -42,14 +42,63 @@ abstract contract Auth {
...
+
+            if iszero(eq(cachedAuthority, 0)) {
+                // We'll use 4 + 32 * 3 bytes.
+                let callDataLength := 100
+
+                // Get a pointer to some free memory.
+                let freeMemoryPointer := mload(0x40)
+
+                // Update the free memory pointer for safety.
+                mstore(0x40, add(freeMemoryPointer, callDataLength))
```
transmissions11, line 57: Do we need to do this? Won't the compiler safely clean before writing to the slot again?

transmissions11:
ideally we can fuzz an Auth child contract to ensure isAuthorized NEVER reverts. it should always return true so an owner can access the contract if the authority breaks

PR #45, opened 2021-10-31
⚡️ Optimize SafeTransferLib

transmissions11:


src/utils/SafeTransferLib.sol
```diff
@@ -26,34 +31,123 @@ library SafeTransferLib {
...
+
+        assembly {
+            // We'll use 4 + 32 * 3 bytes.
+            let callDataLength := 100
+
+            // Get a pointer to some free memory.
+            let freeMemoryPointer := mload(0x40)
+
+            // Update the free memory pointer for safety.
+            mstore(0x40, add(freeMemoryPointer, callDataLength))
```
transmissions11, line 44: Do we need to do this? Won't the compiler safely clean before writing to the slot again?

src/utils/SafeTransferLib.sol
```diff
@@ -26,34 +31,123 @@ library SafeTransferLib {
...
+        assembly {
+            // We'll use 4 + 32 * 2 bytes.
+            let callDataLength := 68
 
-        require(success && (data.length == 0 || abi.decode(data, (bool))), "TRANSFER_FAILED");
+            // Get a pointer to some free memory.
+            let freeMemoryPointer := mload(0x40)
+
+            // Update the free memory pointer for safety.
+            mstore(0x40, add(freeMemoryPointer, callDataLength))
```
transmissions11, line 74: Do we need to do this? Won't the compiler safely clean before writing to the slot again?

src/utils/SafeTransferLib.sol
```diff
@@ -26,34 +31,123 @@ library SafeTransferLib {
...
+
+        assembly {
+            // We'll use 4 + 32 * 2 bytes.
+            let callDataLength := 68
+
+            // Get a pointer to some free memory.
+            let freeMemoryPointer := mload(0x40)
+
+            // Update the free memory pointer for safety.
+            mstore(0x40, add(freeMemoryPointer, callDataLength))
```
transmissions11, line 103: Do we need to do this? Won't the compiler safely clean before writing to the slot again?

PR #58, opened 2021-11-27
🔧 ERC20 - use dummy balance trick to prevent common user transfer errors

z0r0z:
include elementFi safety trick for tokens sent to null or token itself
https://github.com/element-fi/council/blob/13b02be0f7bab784ea7d0e9f28fbb2ef498cdcfe/contracts/libraries/ERC20Permit.sol#L45-L51
basically what you flagged ser

### DIFF
--- a/src/tokens/ERC20.sol
+++ b/src/tokens/ERC20.sol
@@ -60,6 +60,12 @@ abstract contract ERC20 {
 
         INITIAL_CHAIN_ID = block.chainid;
         INITIAL_DOMAIN_SEPARATOR = computeDomainSeparator();
+        
+        // This prevents common user errors by using 
+        // arithmetic safety checks on dummy balances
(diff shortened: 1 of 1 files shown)


transmissions11:
hmm this is a tough one because it technically breaks some invariants about ERC20s (balance of all users summed = total supply) 

tbh i dont rly care about ppl who send their tokens to burn addresses— but i know there are ppl who do (s/o yall with morals ❤️ )

im leaning on the side of not including, as any contract that inherits ERC20 can impl this just as easily (and since solmate ERC20s are abstract, it's not like inheriting the contract is something off the beaten path anyway)

z0r0z:
I see what you mean. Also, for example, the `ENS` token drop would not have worked if this was included, since tokens were pulled from token contract itself. Will close (and this will help be memo for extensions).

Issue #29, opened 2021-10-01
🔍 Evaluate use of underscore prefix for internal functions


transmissions11:
We currently only use the underscore prefix for the `_mint`/`_burn` functions in ERC20, but ideally we have a policy written down of when we use underscore. 

I think my policy would be only using them if a child contract inheriting a solmate parent would want to name a function an unprefixed version (like in the case of `mint`/`burn`)

PR #61, opened 2021-11-28
🎨 ERC20: Style Nits

z0r0z:
- underscore `computeDomainSeparator` to match format of other internal functions (`_mint` / `_burn`)
- use periods rather than exclamation points to conform format and be more stoic, like a warrior monk ;0

### DIFF
--- a/src/tokens/ERC20.sol
+++ b/src/tokens/ERC20.sol
@@ -59,7 +59,7 @@ abstract contract ERC20 {
         decimals = _decimals;
 
         INITIAL_CHAIN_ID = block.chainid;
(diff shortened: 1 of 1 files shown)


transmissions11:
not a fan of underscores for internal functions. [only used for `_mint` and `_burn` cuz they could clash with children who want to impl public `mint` and/or `burn` funcs](https://github.com/Rari-Capital/solmate/issues/29) 😅

transmissions11:
to make that clear maybe we should rename them from `_mint` and `_burn` to something like `print` and `destroy`

transmissions11:
maybe gettin a little too maker-y tho hehe

PR #64, opened 2021-12-11
ERC20vote 🗳️

z0r0z:
add solmate-styled, optimized version of COMP voting token


transmissions11:
lil bit too opinionated for solmate imo— would rather point ppl to kali

z0r0z:
word :~)

PR #74, opened 2021-12-16
📦 Base64 Encoding and Decoding Lib 📚 

abigger87:


transmissions11:
hmmm not sure i want to include this rn actually, lot of audit surface and not used much outside of generative nfts

Vectorized:
Just for fun:

https://gist.github.com/Vectorized/e4b5736fdd24c32fb98b790f1bb36bfe 

Saves 1k gas for short inputs like 'test12' over the current famous implementation. Not that people will call it on-chain though.

transmissions11:
actually pretty cool @Vectorized wanna open a PR?

PR #63, opened 2021-12-11
ERC721 ✨✨

z0r0z:
adds solmate-styled NFT with more contemporary changes

PR #75, opened 2021-12-17
✨ ERC721

distractedm1nd:
Reposting from suggestion in #63: 

Hey, the tests were not running at all, not only because they were set to symbolic execution, but also because much of the ERC20 logic wasn't correctly mapped onto ERC721. Now they work, but I'm not sure if they are adequate to pull into the main branch. I still have some concerns that I wanted to look into over the weekend, mainly about the effects of transfers to address(0).

Also, I changed the tokenURI functionality to match what most projects use today - concatenating a url to a tokenID. This saves us the massive storage that the map would have taken up and just gives us a view function that whatever frontend calls to resolve the metadata

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # solmate
 
-**Modern**, **opinionated** and **gas optimized** building blocks for **smart contract development**.
+**Modern**, **opinionated**, and **gas optimized** building blocks for **smart contract development**.
 
 ## Contracts
 
--- a/src/auth/authorities/RolesAuthority.sol
+++ b/src/auth/authorities/RolesAuthority.sol
@@ -93,12 +93,14 @@ contract RolesAuthority is Auth, Authority {
         bytes4 functionSig,
         bool enabled
     ) public virtual requiresAuth {
-        bytes32 lastRoles = getRoleCapabilities[target][functionSig];
+        bytes32 lastCapabilities = getRoleCapabilities[target][functionSig];
 
         unchecked {
             bytes32 shifted = bytes32(uint256(uint256(2)**uint256(role)));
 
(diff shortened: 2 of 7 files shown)


z0r0z (commented):
this is awesome, and appreciate wrestling through tests.

I hesitate to make tokenURI opinionated to baseURI format, as those tend to be references to centralized dbs? I would also vertically stack the vars in constructor, and make vars in `_mint` horizontal to fit style.

distractedm1nd:
Thanks!

is not a reference to a centralized db, because most (legit) projects upload a folder to IPFS - then it works just as intended. `ipfs://hashOfEntireFolderContents/1` would resolve correctly for example. BUT I noticed a problem: I am not using something like `Strings for uint256`, so concatting like that won't work - going to fix

Also thanks for the heads up about the constructor, didn't notice

alephao:
So the `transferFrom` tests were failing for some cases, I remade them, can you guys have a look?  https://github.com/distractedm1nd/solmate/pull/1/files

Also added test two edge cases: minting token with same id and burning inexistent tokens

I feel there are still a bunch of non-happy path tests missing, will try looking further when I get some more time

alephao:
> I hesitate to make tokenURI opinionated to baseURI format, as those tend to be references to centralized dbs?

What do you suggest? The only alternative I can think of right now is to make the contract abstract with an empty implementation of the `tokenURI` method, thoughts?

distractedm1nd:
[Directories on IPFS are just as immutable as single files](https://medium.com/infinitism/erc-4337-account-abstraction-without-ethereum-protocol-changes-d75c9d94dc4a). How would this be centralized?

transmissions11:
can you take this out of draft i'll merge into `audit-fixes` and productionize from there thank you all for your great work ❤️

transmissions11:
oh woah i can do that myself nice

PR #77, opened 2021-12-19
✨ V6: First Production Release

transmissions11:


transmissions11:
bound() does not satisfy its core invariants 

`bound(0, 68, 69) = 70`
`bound(0, 115792089237316195423570985008687907853269984665640564039457584007913129639654, 115792089237316195423570985008687907853269984665640564039457584007913129639678) = 115792089237316195423570985008687907853269984665640564039457584007913129639675`

transmissions11:
> bound() does not satisfy its core invariants

i think i fixed this in the commit right above

transmissions11:
is permit a thing that goes in 721 and 1155? gotta do some research

distractedm1nd:
generally no, it's nice to have but it's not typical and the vast majority of projects would never use, especially cause most nft projects are on l1 where relaying wouldn't be a thing anyways

including it would depend I guess on what solmate is trying to be?

transmissions11:
ok im not gonna include it cuz seems like there's diverging ideas on how it should be impl'd (uni uses some weird draft eip in theirs)

PR #78, opened 2021-12-19
♻️ Remove early fuzz returns by wrapping inputs

mds1:
Changes:
- Adds `wrap` method to `DSTestPlus` to bound fuzzer inputs
- Removes early returns from tests that fuzz uints by (1) using the `wrap` method, and (2) shrinking inputs that cause overflow
- Add `cache` folder to gitignore (generated by Foundry)

All modified tests were fuzzed with 10k runs and passed locally.

Figuring out what emojis to use on my commit messages was the hardest part of this PR. If you advise on emojis I'll update the commit messages 😛


transmissions11:
thx matt!

PR #80, opened 2021-12-21
 Use a sequential tokenId for erc721 minting

alcueca:
Minting fights are fun, but ultimately wasteful.

### DIFF
--- a/src/tokens/ERC721.sol
+++ b/src/tokens/ERC721.sol
@@ -28,7 +28,9 @@ abstract contract ERC721 {
                             ERC-721 STORAGE
     //////////////////////////////////////////////////////////////*/
     
-    uint256 public totalSupply;
+    uint256 public minted;
+
+    uint256 public burned;
     
     mapping(address => uint256) public balanceOf;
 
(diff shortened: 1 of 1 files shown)


alcueca:
I probably broke your tests, sorry. I would need to find out how to fix the imports with dapptools.

transmissions11:
lol nw they dont work anyway rn i plan to fix today

src/tokens/ERC721.sol
```diff
@@ -28,7 +28,9 @@ abstract contract ERC721 {
                             ERC-721 STORAGE
     //////////////////////////////////////////////////////////////*/
     
-    uint256 public totalSupply;
+    uint256 public minted;
+
+    uint256 public burned;
```
transmissions11, line 33: why use two seperate vars here? think i'd prefer to keep one totalSupply but willing to have my mind changed

src/tokens/ERC721.sol
```diff
@@ -273,16 +282,13 @@ abstract contract ERC721 {
...
-  
-        // Cannot realistically overflow from incrementing total supply beyond
+        address to
+    ) internal virtual returns (uint256 tokenId) { 
+        // Cannot realistically overflow from minting beyond
         // the max uint256 value, and because the sum of all user balances 
         // can't exceed the max uint256 value.
         unchecked {
-            totalSupply++;
+            tokenId = minted++;
```
wilsoncusack, line 291: Personally think should be idiomatic to start at 1 rather than 0 🙂
```suggestion
            tokenId = ++minted;
```

transmissions11: TIL about that syntax woah

transmissions11:
this is neat but i think it's a bit too opinionated for a generic lib— users can always make their mints sequential using their own `nextID` variable or similar like uniswap: https://github.com/Uniswap/v3-periphery/blob/master/contracts/NonfungiblePositionManager.sol#L64

PR #84, opened 2021-12-24
⚡️ Rewrite FixedPointMathLib sqrt() in Yul

ZeframLou:


src/utils/FixedPointMathLib.sol
```diff
@@ -120,56 +120,88 @@ library FixedPointMathLib {
     //////////////////////////////////////////////////////////////*/
 
     function sqrt(uint256 x) internal pure returns (uint256 result) {
-        if (x == 0) return 0;
+        assembly {
+            switch x
```
transmissions11, line 124: i wonder if a not zero check would be cheaper than an switch?

transmissions11:
reminder to self: gotta port over the comments from `audit-fixes` too

src/utils/FixedPointMathLib.sol
ZeframLou: true true, one sec

transmissions11: wow wtf that saved a lot more than i was expecting, i wanna take a look in the hevm debugger in the morning

transmissions11:
also could we just do if x or is the double iszero() necessary 👁️

transmissions11:
> also could we just do if x or is the double iszero() necessary 👁️

checked, optimizer does it for us.

PR #85, opened 2021-12-24
⚡️ Further optimize fpow

ZeframLou:


transmissions11:
huh interesting, is this still cheaper than `and` with the arg order flipped?

PR #86, opened 2021-12-24
⚡️ Further optimize fdiv

ZeframLou:


transmissions11:
i wonder if this would be cheaper if we used the nested if trick like you do in #85

PR #88, opened 2021-12-27
ERC4626 Vault

Joeysantoro:
Core Vaults logic as a distinct helper class for solmate


src/tokens/ERC20Vault.sol
```diff
@@ -0,0 +1,161 @@
...
+
+    /// @notice Creates a new Vault that accepts a specific underlying token.
+    /// @param _underlying The ERC20 compliant token the Vault should accept.
+    /// @param _namePrefix A name prefix to be applied before
+    /// @param _symbolPrefix The ERC20 compliant token the Vault should accept.
+
+    constructor(
+        ERC20 _underlying,
+        string memory _namePrefix,
+        string memory _symbolPrefix
```
transmissions11, line 33: not logic we should enshrine here imo, just take name and symbol

src/tokens/ERC20Vault.sol
```diff
@@ -0,0 +1,161 @@
...
+            _underlying.decimals()
+        )
+    {
+        underlying = _underlying;
+
+        baseUnit = 10**decimals;
+
+        // Prevent minting of shares until
+        // the initialize function is called.
+        totalSupply = type(uint256).max;
```
transmissions11, line 47: there's no initialize here should prolly remove this

src/tokens/ERC20Vault.sol
```diff
@@ -0,0 +1,161 @@
...
+    event Withdraw(address indexed owner, address indexed to, uint256 underlyingAmount);
+
+    /// @notice Deposit a specific amount of underlying tokens.
+    /// @param to The address to receive shares corresponding to the deposit
+    /// @param underlyingAmount The amount of the underlying token to deposit.
+    function deposit(address to, uint256 underlyingAmount) external virtual returns (uint256 shares) {
+        // We don't allow depositing 0 to prevent emitting a useless event.
+        require(underlyingAmount != 0, "AMOUNT_CANNOT_BE_ZERO");
+
+        shares = underlyingAmount.fdiv(exchangeRate(), baseUnit);
```
transmissions11, line 72: lez do this inline like `_mint(to, shares = underlyingAmount.fdiv(exchangeRate(), baseUnit));`

src/tokens/ERC20Vault.sol
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    /// @notice Withdraw a specific amount of underlying tokens.
+    /// @param to The address to receive underlying tokens corresponding to the withdrawal.
+    /// @param underlyingAmount The amount of underlying tokens to withdraw.
+    function withdraw(address to, uint256 underlyingAmount) external virtual returns (uint256 shares) {
+        // We don't allow withdrawing 0 to prevent emitting a useless event.
+        require(underlyingAmount != 0, "AMOUNT_CANNOT_BE_ZERO");
+
+        shares = underlyingAmount.fdiv(exchangeRate(), baseUnit);
```
transmissions11, line 92: see comment above

src/tokens/ERC20Vault.sol
```diff
@@ -0,0 +1,161 @@
...
+        // Determine the equivalent amount of shares and mint them.
+        _mint(to, shares);
+
+        emit Deposit(to, underlyingAmount);
+
+        // Transfer in underlying tokens from the user.
+        // This will revert if the user does not have the amount specified.
+        underlying.safeTransferFrom(msg.sender, address(this), underlyingAmount);
+
+        afterDeposit(underlyingAmount);
```
transmissions11, line 82: worries me doing this after an external call

src/tokens/ERC20Vault.sol
```diff
@@ -0,0 +1,161 @@
...
+        // If there are no shares in circulation, return an exchange rate of 1:1.
+        if (shareSupply == 0) return baseUnit;
+
+        // Calculate the exchange rate by dividing the total holdings by the share supply.
+        return totalHoldings().fdiv(shareSupply, baseUnit);
+    }
+
+    /// @notice Calculates the total amount of underlying tokens the Vault holds.
+    /// @return totalUnderlyingHeld The total amount of underlying tokens the Vault holds.
+    function totalHoldings() public view virtual returns (uint256 totalUnderlyingHeld) {
```
transmissions11, line 158: not a fan of named return args for things that just use return

src/tokens/ERC20Vault.sol
StErMi: just out of curiosity (I'm totally new to optimizations and I'm lurking hard on your Twitter timeline :D): with the inline version, you are losing readability by a lot. inlining things let you gain optimization or it's just more of a "style"?

Am I being to naive to think that the bytecode produced is optimized automatically with the 2 line version?

transmissions11: was thinking about style here, tho in some cases iirc i've found inlining cheaper

src/tokens/ERC20Vault.sol
```diff
@@ -0,0 +1,161 @@
...
+
+        afterDeposit(underlyingAmount);
+    }
+
+    /// @notice Withdraw a specific amount of underlying tokens.
+    /// @param to The address to receive underlying tokens corresponding to the withdrawal.
+    /// @param underlyingAmount The amount of underlying tokens to withdraw.
+    function withdraw(address to, uint256 underlyingAmount) external virtual returns (uint256 shares) {
+        // We don't allow withdrawing 0 to prevent emitting a useless event.
+        require(underlyingAmount != 0, "AMOUNT_CANNOT_BE_ZERO");
```
ElliotFriedman, line 90: we should remove this require statement to save gas for users that send a non zero amount.

transmissions11: yup, planned to do in vaults anyway after abdk agreed

src/tokens/ERC20Vault.sol
```diff
@@ -0,0 +1,161 @@
...
+
+        shares = underlyingAmount.fdiv(exchangeRate(), baseUnit);
+        // Determine the equivalent amount of shares and mint them.
+        _mint(to, shares);
+
+        emit Deposit(to, underlyingAmount);
+
+        // Transfer in underlying tokens from the user.
+        // This will revert if the user does not have the amount specified.
+        underlying.safeTransferFrom(msg.sender, address(this), underlyingAmount);
```
ElliotFriedman, line 80: instead of doing a delegatecall out to the safeTransfer library, this code can be inlined to save gas

transmissions11: it doesnt do a delegatecall, internal libraries get inlined

transmissions11: > To realize this in the EVM, code of internal library functions and all functions called from therein will at compile time be included in the calling contract, and a regular JUMP call will be used instead of a DELEGATECALL.

https://docs.soliditylang.org/en/latest/contracts.html#libraries

ElliotFriedman: If this was inlined, you would call it by doing SafeTransferLib.safeTransferFrom. Even then, I don't think it would work, because library functions that change state have to be delegate called:

```Library functions can only be called directly (i.e. without the use of DELEGATECALL) if they do not modify the state (i.e. if they are view or pure functions)```

Comment (transmissions11) [reply src/tokens/ERC20Vault.sol:80]: thats just not true lol— look at it in a debugger

Screen Shot 2021-12-29 at 2 49 58 PM

Screen Shot 2021-12-29 at 2 50 38 PM

Comment (transmissions11) [reply src/tokens/ERC20Vault.sol:80]: another way you would know it was being delegate-called is because you would have to link it. 

i explicitly disable linking in `.dapprc` so its not even possible here

File: src/tokens/ERC20Vault.sol (hunk 3 above)
Comment (Joeysantoro) [reply src/tokens/ERC20Vault.sol:72]: if its a style thing I actually think its more readable outside

File: src/tokens/ERC20Vault.sol (hunk 4 above)
Comment (Joeysantoro) [reply src/tokens/ERC20Vault.sol:92]: same

File: src/tokens/ERC20Vault.sol (hunk 5 above)
Comment (Joeysantoro) [reply src/tokens/ERC20Vault.sol:82]: the intention behind this is that it will probably also do another external call in deposit like staking in the MasterChef or something. I'm down to enforce that to be a separate function but I also like leaving optional hook here

File: src/mixins/yield-erc/ERC20Vault.sol [hunk 9]
```diff
@@ -0,0 +1,151 @@
...
+
+        underlying.safeTransfer(to, underlyingAmount);
+    }
+
+    /// @notice Redeem a specific amount of shares for underlying tokens.
+    /// @param to The address to receive underlying tokens corresponding to the withdrawal.
+    /// @param shareAmount The amount of shares to redeem for underlying tokens.
+    function redeem(address to, uint256 shareAmount) external virtual returns (uint256 underlyingAmount) {
+        // We don't allow redeeming 0 to prevent emitting a useless event.
+        require(shareAmount != 0, "AMOUNT_CANNOT_BE_ZERO");
```
kesar, line 101:
```suggestion
```

perhaps worth to save this gas. user could still spam with an almost 0 worth value using just 1 share. that could be opt-in for developers doing the check in beforeWithdraw() . thoughts?

Joeysantoro: Great suggestion!

src/mixins/yield-erc/ERC20Vault.sol
```diff
@@ -0,0 +1,150 @@
...
+        beforeWithdraw(underlyingAmount);
+
+        underlying.safeTransfer(to, underlyingAmount);
+    }
+
+    /// @notice Redeem a specific amount of shares for underlying tokens.
+    /// @param to The address to receive underlying tokens corresponding to the withdrawal.
+    /// @param shareAmount The amount of shares to redeem for underlying tokens.
+    function redeem(address to, uint256 shareAmount) external virtual returns (uint256 underlyingAmount) {
+        // We don't allow redeeming 0 to prevent emitting a useless event.
```
Joeysantoro, line 100:
```suggestion
```

transmissions11: was brought up by elliot as well fwiw

Screen Shot 2022-01-02 at 10 58 48 AM

src/mixins/yield-erc/ERC20Vault.sol
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    /*///////////////////////////////////////////////////////////////
+                        DEPOSIT/WITHDRAWAL LOGIC
+    //////////////////////////////////////////////////////////////*/
+
+    /// @notice Emitted after a successful deposit.
+    /// @param user The address that deposited into the Vault.
+    /// @param underlyingAmount The amount of underlying tokens that were deposited.
+    event Deposit(address indexed user, uint256 underlyingAmount);
```
transmissions11, line 53: this should have a owner param

src/mixins/yield-erc/ERC20Vault.sol
```diff
@@ -0,0 +1,149 @@
...
+    /// @param underlyingAmount The amount of underlying tokens that were withdrawn.
+    event Withdraw(address indexed owner, address indexed to, uint256 underlyingAmount);
+
+    /// @notice Deposit a specific amount of underlying tokens.
+    /// @param to The address to receive shares corresponding to the deposit
+    /// @param underlyingAmount The amount of the underlying token to deposit.
+    function deposit(address to, uint256 underlyingAmount) external virtual returns (uint256 shares) {
+        shares = underlyingAmount.fdiv(exchangeRate(), baseUnit);
+        // Determine the equivalent amount of shares and mint them.
+        _mint(to, shares);
```
storming0x, line 67: Given this line 

https://github.com/Rari-Capital/solmate/blob/ab1d04a2ba35cd63ebe90d3704ff0a010a576246/src/tokens/ERC20.sol#L181

a deposit to address 0x0 emits an event that looks like a burn to subgraph and other trackers when the supply actually hasnt changed.

See item 1 in audit for yearn vaults, thought it was similar.

https://github.com/yearn/yearn-security/blob/master/audits/20210719_ToB_yearn_vaultsv2/ToB_-_Yearn_Vault_v_2_Smart_Contracts_Audit_Report.pdf

src/mixins/yield-erc/ERC20Vault.sol
```diff
@@ -0,0 +1,149 @@
...
+    /// @return The amount of underlying tokens an share can be redeemed for.
+    function exchangeRate() public view returns (uint256) {
+        // Get the total supply of shares.
+        uint256 shareSupply = totalSupply;
+
+        // If there are no shares in circulation, return an exchange rate of 1:1.
+        if (shareSupply == 0) return baseUnit;
+
+        // Calculate the exchange rate by dividing the total holdings by the share supply.
+        return totalHoldings().fdiv(shareSupply, baseUnit);
```
storming0x, line 141: This may not be a big of an issue, but item 3 in yearn vaults audit may apply.

https://github.com/yearn/yearn-security/blob/master/audits/20210719_ToB_yearn_vaultsv2/ToB_-_Yearn_Vault_v_2_Smart_Contracts_Audit_Report.pdf

First depositor can mint 1 share and then transfer underlying affecting rounding errors depending on base unit?

src/mixins/yield-erc/ERC20Vault.sol
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    /// @notice Calculates the total amount of underlying tokens the Vault holds.
+    /// @return totalUnderlyingHeld The total amount of underlying tokens the Vault holds.
+    function totalHoldings() public view virtual returns (uint256) {
+        return underlying.balanceOf(address(this));
```
storming0x, line 147: This makes exchangeRate like vaults PricePerShare method a state function that can be altered by an external actor by doing direct transfers, so just FYI on using exchangeRate as an oracle for money markets, etc. See cream hack.

transmissions11: yup this is noted in the original Vaults audit from FPS

src/mixins/yield-erc/ERC20Vault.sol
transmissions11: at least in rari we vaults we planned to handle this as uni does in an initialization module, should probably think about how we do this more generally tho, thanks for the reminder!

src/mixins/yield-erc/ERC20Vault.sol
```diff
@@ -0,0 +1,149 @@
...
+        underlying.safeTransfer(to, underlyingAmount);
+    }
+
+    /// @notice Redeem a specific amount of shares for underlying tokens.
+    /// @param to The address to receive underlying tokens corresponding to the withdrawal.
+    /// @param shareAmount The amount of shares to redeem for underlying tokens.
+    function redeem(address to, uint256 shareAmount) external virtual returns (uint256 underlyingAmount) {
+
+        // Determine the equivalent amount of underlying tokens.
+        underlyingAmount = shareAmount.fmul(exchangeRate(), baseUnit);
```
alcueca, line 102: In `deposit` and `withdraw` you define `shares = underlying /↓ exchangeRate`, so here you need to use `underlying = shares *↑ exchangeRate`.

src/mixins/yield-erc/ERC20Vault.sol
```diff
@@ -0,0 +1,149 @@
...
+    /*///////////////////////////////////////////////////////////////
+                        VAULT ACCOUNTING LOGIC
+    //////////////////////////////////////////////////////////////*/
+
+    /// @notice Returns a user's Vault balance in underlying tokens.
+    /// @param user The user to get the underlying balance of.
+    /// @return The user's Vault balance in underlying tokens.
+    function balanceOfUnderlying(address user) external view returns (uint256) {
+        return balanceOf[user].fmul(exchangeRate(), baseUnit);
+    }
```
alcueca, line 129: Same as above, here you should round up.

src/mixins/yield-erc/ERC20Vault.sol
```diff
@@ -0,0 +1,149 @@
...
+    function deposit(address to, uint256 underlyingAmount) external virtual returns (uint256 shares) {
+        shares = underlyingAmount.fdiv(exchangeRate(), baseUnit);
+        // Determine the equivalent amount of shares and mint them.
+        _mint(to, shares);
+
+        emit Deposit(to, underlyingAmount);
+
+        // Transfer in underlying tokens from the user.
+        // This will revert if the user does not have the amount specified.
+        underlying.safeTransferFrom(msg.sender, address(this), underlyingAmount);
```
alcueca, line 73: Consider tracking the balance of underlying in the contract, and here subtract any surplus from the transfer, potentially even not executing the transfer if not needed.

This will allow other contracts that produce or move underlying to leave it at this vault, executing `deposit` afterwards. This will remove the need for an `approve` (less gas) and make integrations easier.

For examples of using this pattern, check the yield contracts.

src/mixins/yield-erc/ERC20Vault.sol
```diff
@@ -0,0 +1,149 @@
...
+
+    /// @notice Withdraw a specific amount of underlying tokens.
+    /// @param to The address to receive underlying tokens corresponding to the withdrawal.
+    /// @param underlyingAmount The amount of underlying tokens to withdraw.
+    function withdraw(address to, uint256 underlyingAmount) external virtual returns (uint256 shares) {
+        shares = underlyingAmount.fdiv(exchangeRate(), baseUnit);
+
+        // Determine the equivalent amount of shares and burn them.
+        // This will revert if the user does not have enough shares.
+        _burn(msg.sender, shares);
```
alcueca, line 86: As above, consider burning any shares held by this contract first. This allows other contracts to move shares (given approval by their owner), saves gas, and simplifies integrations.

src/mixins/yield-erc/ERC20Vault.sol
```diff
@@ -0,0 +1,149 @@
...
+    /// @param to The address to receive underlying tokens corresponding to the withdrawal.
+    /// @param shareAmount The amount of shares to redeem for underlying tokens.
+    function redeem(address to, uint256 shareAmount) external virtual returns (uint256 underlyingAmount) {
+
+        // Determine the equivalent amount of underlying tokens.
+        underlyingAmount = shareAmount.fmul(exchangeRate(), baseUnit);
+
+        // Burn the provided amount of shares.
+        // This will revert if the user does not have enough shares.
+        _burn(msg.sender, shareAmount);
```
alcueca, line 106: As above

alcueca (commented):
Two comments on rounding up on fmul to maintain equivalence to fdiv which rounds down.

Three comments on using tracked balances to facilitate integration and reduce gas costs

src/mixins/yield-erc/ERC20Vault.sol
transmissions11: yup, ABDK also pointed this out in the vaults audit

src/mixins/yield-erc/ERC20Vault.sol
transmissions11: good point

src/mixins/yield-erc/ERC20Vault.sol
transmissions11: this isn't compatible with how we handle fees in the rari vaults (they're minted as shares to the vault itself which an admin can claim)

src/mixins/yield-erc/ERC20Vault.sol
transmissions11: argh good point, would hate to impl a to != 0 check but seems like this might be a valid case

transmissions11: i think we might even want to impl this at the ERC20 level? hmmmmmmm

transmissions11: well wait this is already possible just by transferring shares to address(0). and we're not the only erc20 that supports doing that (see UNIv2ERC20)

src/mixins/yield-erc/ERC20Vault.sol
Elyx0: Just passing by, `an share -> a share`. Great reads around here!

transmissions11: good find @Elyx0 !

src/mixins/yield-erc/ERC20Vault.sol
Joeysantoro: adding

src/mixins/yield-erc/ERC20Vault.sol
Joeysantoro: Not sure if we need to address this, will let @transmissions11 make the final call

src/mixins/yield-erc/ERC20Vault.sol
Joeysantoro: updated the comment

src/mixins/yield-erc/ERC20Vault.sol
Joeysantoro: @transmissions11 can you patch in your fix for vaults?

src/mixins/yield-erc/ERC20Vault.sol
Joeysantoro: I love this suggestion. For Fei PCV deposits we use an "assume tokens were already transferred model"

src/mixins/yield-erc/ERC20Vault.sol
Joeysantoro: @transmissions11 thoughts on adding it here though?

src/mixins/yield-erc/ERC20Vault.sol
```diff
@@ -0,0 +1,149 @@
...
+    /*///////////////////////////////////////////////////////////////
+                                IMMUTABLES
+    //////////////////////////////////////////////////////////////*/
+
+    /// @notice The underlying token the Vault accepts.
+    ERC20 public immutable underlying;
+
+    /// @notice The base unit of the underlying token and hence vault.
+    /// @dev Equal to 10 ** decimals. Used for fixed point arithmetic.
+    uint256 public immutable baseUnit;
```
transmissions11, line 23: this should be internal now that it is no longer enshrined in the EIP

src/mixins/yield-erc/ERC20Vault.sol
```diff
@@ -48,9 +48,10 @@ contract ERC4626Vault is ERC20 {
     //////////////////////////////////////////////////////////////*/
 
     /// @notice Emitted after a successful deposit.
-    /// @param user The address that deposited into the Vault.
+    /// @param sender The address that deposited into the Vault.
+    /// @param to The address that received deposit shares.
     /// @param underlyingAmount The amount of underlying tokens that were deposited.
-    event Deposit(address indexed user, uint256 underlyingAmount);
+    event Deposit(address indexed sender, address indexed to, uint256 underlyingAmount);
```
transmissions11, line 54:
```suggestion
    event Deposit(address indexed from, address indexed to, uint256 underlyingAmount);
```

we call it this in the eip

src/mixins/yield-erc/ERC20Vault.sol
```diff
@@ -0,0 +1,150 @@
...
+    /// @param sender The address that deposited into the Vault.
+    /// @param to The address that received deposit shares.
+    /// @param underlyingAmount The amount of underlying tokens that were deposited.
+    event Deposit(address indexed sender, address indexed to, uint256 underlyingAmount);
+
+    /// @notice Emitted after a successful withdrawal.
+    /// @param owner The address that withdrew from the Vault.
+    /// @param to The destination for withdrawn tokens.
+    /// @param underlyingAmount The amount of underlying tokens that were withdrawn.
+    event Withdraw(address indexed owner, address indexed to, uint256 underlyingAmount);
```
transmissions11, line 60:
```suggestion
    event Withdraw(address indexed from, address indexed to, uint256 underlyingAmount);
```

same as above

src/mixins/yield-erc/ERC20Vault.sol
transmissions11: maybe we make it all caps again as well

src/mixins/yield-erc/ERC20Vault.sol
transmissions11: i need to impl a fmulUp in solmate first, will do in a bit

src/mixins/yield-erc/ERC20Vault.sol
transmissions11: dont want to enshrine that complexity in the defacto "example" impl, but its totally compliant with the EIP, so would love to see someone like Alberto make a skimbatch compatible impl

src/mixins/yield-erc/ERC20Vault.sol
alcueca: I would reverse the exchange rate, use fmul for deposit and withdraw, and fdivup for the rest. I've found it makes for an easier flow to avoid fdiv for as long as possible, and having a consistent approach helps integrations. Personal preference, though.

src/mixins/yield-erc/ERC20Vault.sol
alcueca: I've been banging this drum for a while now. We can use this with just some renaming.

src/mixins/yield-erc/ERC20Vault.sol
StErMi: it's true that it's not a true `constant` but it can be treated as a variant of a constant because after `constructor` act as it. 

For that reason (but it's more a stylistic way to instantly recognize it) it should be uppercase. but this go in the bucket of style choice.

src/mixins/yield-erc/ERC20Vault.sol
```diff
@@ -0,0 +1,150 @@
...
+        if (shareSupply == 0) return baseUnit;
+
+        // Calculate the exchange rate by dividing the total holdings by the share supply.
+        return totalHoldings().fdiv(shareSupply, baseUnit);
+    }
+
+    /// @notice Calculates the total amount of underlying tokens the Vault holds.
+    /// @return totalUnderlyingHeld The total amount of underlying tokens the Vault holds.
+    function totalHoldings() public view virtual returns (uint256) {
+        return underlying.balanceOf(address(this));
```
sircryptomaxi, line 148: There are no checks for balances on the strategy contracts? Since on deposits you pass the underlying to another contract, totalHoldings will return much less than the actual holdings. The exchange rate will be 0 if there’s no underlying left on the vault contract and is fully deployed on strategy contracts.

src/mixins/yield-erc/ERC20Vault.sol
```diff
@@ -0,0 +1,150 @@
...
+
+    constructor(
+        ERC20 _underlying,
+        string memory _name,
+        string memory _symbol
+    )
+        ERC20(
+            _name,
+            _symbol,
+            _underlying.decimals()
```
yaronvel, line 38: why is it necessary to have `baseUnit`? `decimals` is not part of the ERC20 standard.
From personal experience I can say that at least in the years of 2018-2019, some dexes were unable to list tokens because they had `DECIMALS` instead of `decimals` or no decimals at all.

But the real issue is that from what I see there is no real need for base unit here, and it would be better to just have shares and exchange rate denominated with fixed 18 decimals.

src/mixins/yield-erc/ERC20Vault.sol
```diff
@@ -0,0 +1,150 @@
...
+
+        emit Withdraw(msg.sender, to, underlyingAmount);
+
+        // Withdraw from strategies if needed and transfer.
+        beforeWithdraw(underlyingAmount);
+
+        underlying.safeTransfer(to, underlyingAmount);
+    }
+
+    function beforeWithdraw(uint256 underlyingAmount) internal virtual {}
```
yaronvel, line 117: the decision to have only `beforeWithdraw` and `afterDeposit` hooks seems arbitrary.
For example, if one wants to distribute some token reward based on user balance, then it should trigger it before deposit, and not after. (see, e.g., MakerDAO's `cropJoin`).

src/mixins/yield-erc/ERC20Vault.sol
```diff
@@ -0,0 +1,150 @@
...
+    //////////////////////////////////////////////////////////////*/
+
+    /// @notice Emitted after a successful deposit.
+    /// @param sender The address that deposited into the Vault.
+    /// @param to The address that received deposit shares.
+    /// @param underlyingAmount The amount of underlying tokens that were deposited.
+    event Deposit(address indexed from, address indexed to, uint256 underlyingAmount);
+
+    /// @notice Emitted after a successful withdrawal.
+    /// @param owner The address that withdrew from the Vault.
```
kesar, line 57:
```suggestion
    /// @param from The address that withdrew from the Vault.
```

src/mixins/yield-erc/ERC20Vault.sol
```diff
@@ -0,0 +1,150 @@
...
+
+        baseUnit = 10**decimals;
+    }
+
+    /*///////////////////////////////////////////////////////////////
+                        DEPOSIT/WITHDRAWAL LOGIC
+    //////////////////////////////////////////////////////////////*/
+
+    /// @notice Emitted after a successful deposit.
+    /// @param sender The address that deposited into the Vault.
```
kesar, line 51:
```suggestion
    /// @param from The address that deposited into the Vault.
```

src/mixins/yield-erc/ERC20Vault.sol
transmissions11: there are no strategy contracts in this demo implementation, its meant to be overriden

src/mixins/yield-erc/ERC20Vault.sol
belbix: I'm sure need not only before/after hooks but also granulate it for before/after burn/mint and before/after token transfer.
It can be used in fees logic and determination correct vault PPFS in strategies.

src/mixins/yield-erc/ERC20Vault.sol
joshieDo: > This will allow other contracts that produce or move underlying to leave it at this vault, executing deposit **_afterwards_**.

probably self-evident, but this would have to happen atomically right? Otherwise, I can see some sniping happening

src/mixins/yield-erc/ERC20Vault.sol
storming0x: yeah, makes sense to leave as is , given that is understood by users of the lib.

src/mixins/yield-erc/ERC20Vault.sol
Joeysantoro: yeah would be atomic. After discussing with @transmissions11 going to leave it approval only for this base implementation.

src/mixins/yield-erc/ERC20Vault.sol
```diff
@@ -0,0 +1,150 @@
...
+        // If there are no shares in circulation, return an exchange rate of 1:1.
+        if (shareSupply == 0) return baseUnit;
+
+        // Calculate the exchange rate by dividing the total holdings by the share supply.
+        return totalHoldings().fdiv(shareSupply, baseUnit);
+    }
+
+    /// @notice Calculates the total amount of underlying tokens the Vault holds.
+    /// @return totalUnderlyingHeld The total amount of underlying tokens the Vault holds.
+    function totalHoldings() public view virtual returns (uint256) {
```
transmissions11, line 147: thsi should be virtual

src/mixins/yield-erc/ERC20Vault.sol
mosinnik: why _mint going before transfer? It may cause re-entrant attack. So _mint should be after tranferring

PR #89, opened 2021-12-28
♻️ 1155 changes

RagePit:
Optimized the unchecked{} portions of 1155. Included test gas usage above some of the blocks.

safeTransferFrom and safeBatchTransferFrom needs to have the -= as a checked operation to prevent an attacker from underflowing their balance since we aren't verifying the balance implicitly. 

https://github.com/RagePit/solmate-1155 for quick underflow demo

I'm a git and sol noob so would appreciate pointing out mistakes I made


src/test/ERC1155.t.sol
```diff
@@ -14,10 +14,116 @@ contract ERC1155Test is DSTestPlus {
...
+        
+        uint len = ids.length;
+        uint256[] memory amounts = new uint256[](len);
+        address[] memory owners = new address[](len);
+
+        for (uint i = 0; i < len; i++) {
+            //if ids array has any dupes amounts will not be correct
+            for (uint256 j = 0; j < i; j++) {
+                if (ids[i] == ids[j]) return;
+            }
```
RagePit, line 38: These shit for fuzzing times but will fail with dupes otherwise. Could use a better test solution or just ignore the fails

transmissions11:
thx!

Issue #93, opened 2022-01-02
ERC721 proxy approval


cxkoda:
The current implementation of `ERC721.transferFrom` is hardwired to check storage for `isApprovedForAll`
https://github.com/Rari-Capital/solmate/blob/b150e10ec0c2e6ea5fb05a7284cdbf6d9e449427/src/tokens/ERC721.sol#L79

This prevents e.g. additional contract-side pre-approval of proxy registries, such as the [WyvernProxy](https://etherscan.io/address/wyvernproxyregistry.eth) by OpenSea, which might be desirable in some cases.

transmissions11:
feel free to override transferFrom or fork, but this is not something we plan to virtualize at the isApprovedForAll level

cxkoda:
I see. Thanks for the feedback!

Just out of curiosity: Is there a technical reasoning against this? Or is just not compatible with the goal for simplicity?

transmissions11:
mostly for the aesthetic of simplicity yeah, tho as seen in your PR calling functions instead of reading straight from storage adds extra JUMPs in the bytecode which increases gas consumption 

<img width="1395" alt="Screen Shot 2022-01-02 at 2 59 27 PM" src="https://user-images.githubusercontent.com/26209401/147891684-1df4b5f6-5ac2-4bea-9ab4-fccaa7f953a4.png">

cxkoda:
I see. I've been a little puzzle about this, because I also expected the gas costs to be increased due to the indirection. However, in 3 out of 4 cases the gas consumption is actually `15` lower than before.

transmissions11:
its due to the way function dispatch is changed by adding a public function instead of using a public state var. solidity is super weird with how thats fetched, changing one tiny function name or location can change the lookup order of all functions, so calling one might be slightly cheaper at the cost of another

cxkoda:
ah yes, that makes sense. since I introduce `isApprovedForAll` later, the order in the lookup will surely change. thanks!

PR #94, opened 2022-01-02
Virtualize isApprovedForAll method for ERC721

cxkoda:
Possible solution for #93


transmissions11:
dont plan to support this kind of modularity in solmate contracts, would recommend forking (or overriding transferFrom)

PR #96, opened 2022-01-03
Optimize erc20 for transfer operation [WIP]

yaronvel:
Standard solidity compilation will not put `transfer` as the first function selection.
This PR implements an ERC20 that only has a fallback function and no other public functions.
When calling the contract the fallback function checks if it is a call to `transfer` and implements the transfer logic.
Otherwise it delegates the call to an impl contract (this cost more gas, but called less often).

Gas improvement is between 0.5% to 2%, depending on the scenario (e.g., transfer to an account that already holds the token or to a fresh one. call from a tx or an internal call from another contract, etc).

If you want to go this line of optimizations, can also add support to `transferFrom`, which is also commonly used.
But I really hope you won't go that low :)

### DIFF
--- a/src/tokens/OptERC20.sol
+++ b/src/tokens/OptERC20.sol
@@ -1,0 +1,101 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity >=0.8.0;
+
+contract OptERC20 {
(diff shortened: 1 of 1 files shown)


transmissions11:
hehe a bit too overcomplicated for me but i appreciate the pr

PR #99, opened 2022-01-05
🦺 Fix ERC1155 fuzz tests

brockelmore:
If minLength is 0, it bypasses the for loop and directly calls, which doesn't fail.


transmissions11:
closes #100

transmissions11:
i think a better condition is checking if all the arrays are length 0, in the case that only one or two of them is this should still fail, i like being precise (no need to make this change urself ill do tmrw)

PR #108, opened 2022-01-13
spec fixes

Joeysantoro:
Fixes withdraw/redeem to conform to most updated spec
Adds mint

fixes bug in calculateShares

### DIFF
--- a/src/mixins/ERC4626.sol
+++ b/src/mixins/ERC4626.sol
@@ -90,9 +90,11 @@ abstract contract ERC4626 is ERC20 {
             allowance[from][msg.sender] -= shareAmount;
         }
 
+        underlyingAmount = calculateUnderlying(shareAmount);
+
         _burn(from, shareAmount);
 
-        emit Withdraw(from, to, underlyingAmount = calculateUnderlying(shareAmount));
+        emit Withdraw(from, to, underlyingAmount);
 
         beforeWithdraw(underlyingAmount);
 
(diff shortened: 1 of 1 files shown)


src/mixins/ERC4626.sol
```diff
@@ -59,48 +59,44 @@ abstract contract ERC4626 is ERC20 {
         afterDeposit(underlyingAmount);
     }
 
-    function withdraw(address to, uint256 underlyingAmount) public virtual returns (uint256 shares) {
-        _burn(msg.sender, shares = calculateShares(underlyingAmount));
+    function mint(address to, uint256 shareAmount) public virtual returns (uint256 underlyingAmount) {
+        _mint(to, shareAmount);
 
-        emit Withdraw(msg.sender, to, underlyingAmount);
+        emit Deposit(msg.sender, to, underlyingAmount = calculateUnderlying(shareAmount));
```
zerosnacks, line 65: This is more of a code style preference but I would prefer it if `underlyingAmount = calculateUnderlying(shareAmount)` was defined outside of the `emit`, it is more explicit in my opinion seeing as we are using the underlyingAmount in multiple places after the event as well.

src/mixins/ERC4626.sol
```diff
@@ -59,48 +59,44 @@ abstract contract ERC4626 is ERC20 {
         afterDeposit(underlyingAmount);
     }
 
-    function withdraw(address to, uint256 underlyingAmount) public virtual returns (uint256 shares) {
-        _burn(msg.sender, shares = calculateShares(underlyingAmount));
+    function mint(address to, uint256 shareAmount) public virtual returns (uint256 underlyingAmount) {
+        _mint(to, shareAmount);
```
zerosnacks, line 63: I think this is incorrect, this should be `_mint(to, underlyingAmount = calculateUnderlying(shareAmount)`.
In the current implementation it always returns an `underlyingAmount` of 0.

src/mixins/ERC4626.sol
zerosnacks: (see comment below, this is actually part of a bug I think)

zerosnacks:
LGTM 👍

Joeysantoro:
@transmissions11 closing because this landed in solmate-next repo. We should double check and delete this branch to prevent confusion

PR #110, opened 2022-01-18
Fixed ERC4646 link

StErMi:
given that the contract would be merged with he current folder structure the current version would point to `mixinz` folder but the contract is stored in `mixins`


transmissions11:
good find thanks chief

transmissions11:
done in master

PR #119, opened 2022-01-31
add max functions

Joeysantoro:


transmissions11:
Ordering of events is wrong, and would like to check the limits in the functions so we can easily override to add that functionality

PR #121, opened 2022-02-01
Fix/withdraw allowance

Ratimon:
In the `withdraw`  function:

This PR points that **shares** variable has not been initialised as  `previewWithdraw(amount)` when modifying **allowance** state. So, The allowance state of `from` could be possibly wrong  :)


transmissions11:
Good find, remove the solmate import and I’ll merge

transmissions11:
Fixed manually on main, sorry to frontrun but this is pretty critical

PR #125, opened 2022-02-10
No need to initialise loop increment variable with default value

pcaversaccio:
If a variable is not set/initialised, it is assumed to have the default value. For `uint` it's 0. If you explicitly initialise it with its default value, you are just wasting gas.

Also, I refactored the unchecked increment code. The function `_uncheckedInc` will be inlined for Solidity versions starting from `0.8.2`.


transmissions11:
it doesn't waste gas to set a variable to its default value, and i'm not a fan of _uncheckedInc for stylistic reasons, but thanks for your pr!

PR #126, opened 2022-02-11
✨ More safeCasts

Joeysantoro:


transmissions11:
makes sense, can you add tests?

Joeysantoro:
> makes sense, can you add tests?

done king

PR #128, opened 2022-02-11
✨ added EIP712 util contract

naomsa:
Added a simple but powerful eip712 util contract based on assembly processing and inherited it on ERC20.sol to have a better workflow on the permit feature


naomsa:
just realized that the addition of the contract into erc20 increased gas by a small amount (~200 units) so we could revert that commit

transmissions11:
not something i'd like to add

PR #129, opened 2022-02-14
simplify muldiv requires

ZumZoom:


transmissions11:
ayyy thanks for the PR, sadly this seems to have the opposite effect and actually increases gas costs 😅 

Screen Shot 2022-02-15 at 8 01 53 PM

k06a:
BTW this assembly code is not equivalent to the Solidity code,  but similar. Because Solidity always has [Short-circuit evaluation](https://en.wikipedia.org/wiki/Short-circuit_evaluation) of the boolean expressions.

k06a:
Agree, I would fix comment a bit to make it more clear that code is not completely equivalent

PR #134, opened 2022-02-18
🔒 Fix ERC4626 mint() shares amount

eswak:
The amount minted is the amount of assets, not the amount of shares.

Also update the test file to validate the issue. It was not detected because no deposit/mint were done after the vault had accrued yield.

Also add a test to validate that we can mint/deposit/withdraw/redeem for other addresses, as all tests mint shares to themselves.


transmissions11:
great find, chad as always

PR #135, opened 2022-02-23
⚡ Save MLOAD in ERC1155

z0r0z:
use ownersLength to save mload


transmissions11:
oo good find not sure how i missed that

PR #138, opened 2022-02-26
ERC721A Solmate implementation for cheaper batch minting

nuhhtyy:
ERC721A is a useful standard for cheaper batch minting that cost similar to the price of a single ERC721 mint with minimal tradeoffs for end users


transmissions11:
not something i'd like to add

PR #139, opened 2022-02-28
⚡️ replace require by custom errors to save gas on deployment

qd-qd:
## Context

Starting from Solidity [v0.8.4](https://blog.soliditylang.org/2021/04/21/custom-errors/), there is a convenient and gas-efficient way to explain to users why an operation failed through the use of custom errors. This way is less expensive, especially when it comes to deploying cost than passing a string to require or assert.

### Before

Screenshot 2022-03-02 at 15 25 47

### After

Screenshot 2022-03-02 at 15 25 56

## Tradeoff

This improvement requires a bump to 0.8.**4**.  I am personally convinced this is worth it and the accessibility issue caused by the bump is acceptable due to the fact the bump is minor.

--- 

What do you think about bumping the contracts to 0.8.4?
If you are okay with the changes, I could replace all `require()` with custom errors.

### DIFF
--- a/src/tokens/ERC721.sol
+++ b/src/tokens/ERC721.sol
@@ -1,5 +1,15 @@
 // SPDX-License-Identifier: AGPL-3.0-only
-pragma solidity >=0.8.0;
+pragma solidity >=0.8.4;
+
+/*///////////////////////////////////////////////////////////////
+                             ERRORS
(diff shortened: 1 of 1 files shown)


qd-qd:
added gas reporting ✅

transmissions11:
don't plan to add yet, sorry!

PR #143, opened 2022-03-02
💥 ERC4626 interface updates

### LINKED ISSUE #142: Update ERC4626 Interface
https://github.com/Rari-Capital/solmate/blob/7f433c79c498f7cef10e1144e617f3359d19a4b4/src/mixins/ERC4626.sol#L123

`assetsOf` and `assetsPerShare` are no longer in the interface (now in last call)

these should be changed to `convertToShares` and `convertToAssets`

eswak:
ERC is now in last call.

Use proper wording for variables:
- amount -> assets
- from/user -> owner
- to -> receiver

remove assetsOf and assetsPerShare

add convertToShares and covnertToAssets

previewMint, previewDeposit, previewRedeem, previewWithdraw are not read-only anymore

change asset() type to address instead of ERC20, to use 100% standard typings on external interfaces

update allowance management in redeem to be identical to withdraw

Closes #142


eswak:
good catch

src/mixins/ERC4626.sol
```diff
@@ -120,33 +122,37 @@ abstract contract ERC4626 is ERC20 {
 
     function totalAssets() public view virtual returns (uint256);
 
-    function assetsOf(address user) public view virtual returns (uint256) {
-        return previewRedeem(balanceOf[user]);
+    function convertToShares(uint256 assets) public view returns (uint256 shares) {
+        uint256 supply = totalSupply; // Saves an extra SLOAD if totalSupply is non-zero.
+
+        return shares = supply == 0 ? assets : assets.mulDivUp(supply, totalAssets());
```
Joeysantoro, line 128: this should mulDivDown not up to track the results of a deposit call imo. Open to tracking withdraw if thats preferred by @transmissions11

transmissions11: I’d like to standardize this in the spec itself, can see arguments for both sides. Needs a bit more discussion

transmissions11: oke we agreed to use mulDivDown

eswak: I'll push the change, but just a heads up, this assertion won't hold : 

`vault.convertToShares(vault.convertToAssets(vault.balanceOf(x))) !== vault.balanceOf(x)`

because both functions round down (pessimisticly)

transmissions11: 0B9AF509-3F9C-4554-8581-CA60C8D3A197

Even if we rounded up its still not reversible in all cases tho, no?

Joeysantoro: I don't think so either

PR #144, opened 2022-03-02
⚡️ take advantage of the low-level inliner

qd-qd:
## Context

The [v0.8.2](https://blog.soliditylang.org/2021/04/21/custom-errors/) adds a simple inliner to the low-level optimizer of Solidity. It can inline short functions that do not contain control-flow branches or opcodes with side-effects.

### Before

Screenshot 2022-03-02 at 15 48 25

### After

Screenshot 2022-03-02 at 15 48 33

## Note

Would be happy to discuss this with you guys if I miss something here. I assume that the logic was duplicated to avoid dealing with `jump` opcode etc.. but if I understand correctly, this is now managed by the low-level inliner.

### DIFF
--- a/src/tokens/ERC721.sol
+++ b/src/tokens/ERC721.sol
@@ -100,14 +100,7 @@ abstract contract ERC721 {
         address to,
         uint256 id
     ) public virtual {
-        transferFrom(from, to, id);
-
-        require(
-            to.code.length == 0 ||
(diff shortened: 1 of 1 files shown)


transmissions11:
Screen Shot 2022-03-03 at 8 45 28 PM
tested with 0.8.10, raises gas across the board

PR #145, opened 2022-03-02
ERC721: read from `calldata` for `safeTransferFrom()` and `_safeMint()`

z0r0z:


transmissions11:
hmmm i kept memory for flexibility when used internally

PR #148, opened 2022-03-06
⚡️ Inline variable in ERC20 permit

outdoteth:
The digest is only used once. I haven't actually looked at the bytecode output but I'm quite sure that by inlining instead of declaring a variable an MSTORE and MLOAD is saved.

Before cost:

Screenshot 2022-03-06 at 11 19 23

After cost:

Screenshot 2022-03-06 at 11 19 50

Savings: ~32 gas


outdoteth:
Hmm, @transmissions11   after your comment on the other PR I am now not so sure if an MLOAD or MSTORE is the cause of the savings here. A bytes32 should be on the stack shouldn't it? Will take a look at the assembly tonight.

transmissions11:
hashing is a bit different since the stack input gets lifted into memory before being hashed, but i believe this will still just save some stack operations (its a good opti tho, ill consider it)

outdoteth:
cool, yep I took a look and it's saving a few swaps and a couple dups.

PR #149, opened 2022-03-06
⚡️ Reorder variables in ERC1155

outdoteth:
Memory is append-only in solidity ([ref](https://docs.soliditylang.org/en/v0.8.12/internals/layout_in_memory.html)).

> "Solidity always places new objects at the free memory pointer and memory is never freed (this might change in the future)."

By initialising and modifying the intermediate variable before the loop, you can avoid repeatedly pushing to memory. Gas saving will be `base savings * n` since this changes the cost from O(n) to O(1).

Before cost:
Screenshot 2022-03-06 at 12 02 15

After cost:
Screenshot 2022-03-06 at 12 02 21

From looking at the tests I think the gas saving is approx: 
~20 gas per token


transmissions11:
thanks for your PR! this indeed saves gas, but just fyi these variables aren't stored in memory, they live on the stack. this saves gas because the compiler is able to use a more efficient series of stack operations to do the same task:


Screen Shot 2022-03-06 at 11 44 48 AM

outdoteth:
Ohh cool. I shouldn't have assumed. That was just a lucky change then 😄 .

PR #151, opened 2022-03-06
✨ Merkle Lib

rajivpo:
should be more gas efficient than the [OZ implementation](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/MerkleProof.sol) which was used as a starting point.

First time writing something only in Yul so any and all comments/suggestions are welcome 🙂

### DIFF
--- a/src/utils/MerkleProof.sol
+++ b/src/utils/MerkleProof.sol
@@ -12,20 +12,13 @@ library MerkleProof {
     ) internal pure returns (bool isValid) {
         assembly {
             let computedHash := leaf
-            // The first slot of a dynamically sized array stores the array length.
-            let proofLength := mload(proof)
-
-            // Exit early if an empty proof is supplied.
-            if iszero(proofLength) {
-                revert(0, 0)
-            }
 
             // Get the memory start location of the first element in the proof array.
             let data := add(proof, 0x20)
 
             // Iterate over proof elements to compute root hash.
             for {
-                let end := add(data, mul(proofLength, 0x20))
+                let end := add(data, mul(mload(proof), 0x20))
             } lt(data, end) {
                 data := add(data, 0x20)
             } {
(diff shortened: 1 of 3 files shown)


src/utils/MerkleProof.sol
```diff
@@ -0,0 +1,40 @@
...
+        bytes32[] memory proof,
+        bytes32 root,
+        bytes32 leaf
+    ) internal pure returns (bool isValid) {
+         assembly {
+            let computedHash := leaf
+            let proofLength := mload(proof)
+
+            // exit early if empty proof is supplied
+            if iszero(proofLength) {
```
itsdevbear, line 15: probably can remove this, adds gas for the success case.

transmissions11:
this is dope! thanks @rajivpo 

would you mind running prettier and giving this some comments in the style of our other yul libs?

src/utils/MerkleProof.sol
rajivpo: this is true - @transmissions11 what do you think?

transmissions11: As long as it still reverts in other cases I agree

transmissions11:
Looking better! Do we have an estimate of how much gas this saves vs OZ? Would also like some fuzz tests… i believe there are some other repos we can reference for generating trees in solidity

rajivpo:
> Looking better! Do we have an estimate of how much gas this saves vs OZ? Would also like some fuzz tests… i believe there are some other repos we can reference for generating trees in solidity

image

Can definitely add some fuzz tests too.

transmissions11:
bump on fuzz tests, have a great use case for this 😉

rajivpo:
> bump on fuzz tests, have a great use case for this 😉

ye will try to wrap up tomorrow

transmissions11:
lfg

PR #152, opened 2022-03-09
♻️ Reorder ERC721 approval check

ahbanavi:
Following the changes implemented in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3248


transmissions11:
I mean do we know if using specific approval is less common then all? This isn't a blanket optimization, it's a trade off

transmissions11:
i dont keep up with the nft industry much, but i'll take ur word that approvealls are more common than individual approvals...

PR #156, opened 2022-03-11
♻️ SafeCastLib - Gas reduction, and more types

0xClandestine:
Gas was reduced by replacing require() with if() statements, and setting constants for type(UINT#).max


transmissions11:
this saves no gas and makes things ugly imo

feel free to reopen if u just want o add more types, but don't base it off the rsqrt branch pls

PR #165, opened 2022-03-18
⚡️ Optimize SafeCastLib

134dd3v:
## Description

EVM instruction set does not have "less than or eq" opcode, it just has "less than" and "equal". This PR simply updates the logic to use "less than".

## Checklist

- [x] Ran `dapp snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?
- [x] Ran `dapp test`?


transmissions11:
oh nice great idea!

PR #171, opened 2022-03-21
✨ FixedMathLib: expWadDown()

recmo:
## Description

Describe the changes made in your pull request here.

## Checklist

- [ ] Ran `dapp snapshot`?
- [ ] Ran `npm run lint`?
- [x] Ran `forge test`?
- [ ] Ran `dapp test`?


recmo:
Ping @transmissions11

transmissions11:
👀👀👀👀👀👀👀👀👀

transmissions11:
amazing already

Screen Shot 2022-03-21 at 11 22 47 PM

recmo:
Hmm... that should have been a ~10 gas reduction.

transmissions11:
well if we inline everything into q should save?

recmo:
Can't inline the variables `z` and `w` because they both get used twice. That's the scheme basically, you need more variables and additions, but in return you have fewer multiplies. Multiplies are more expensive, so it may be worthwhile.

transmissions11:
yup... thought maybe this would save some but looks like the optimizer was smart enough in the first place

Screen Shot 2022-03-21 at 11 47 08 PM

recmo:
We can always go back to the original formula for `q`, but somehow I feel that removing a multiply should save us something :D. 

I have a similar scheme for `p` that also removes a mul:

```
int256 y = x      +     1346386616545796478920950773328;
y = (y * x >> 96) +    57155421227552351082224309758442;
int256 p = y + x  -    94201549194550492254356042504812;
p = (p * y >> 96) + 28719021644029726153956944680412240;
```

but that one somehow ends up costing 1 gas more than the plain formula in my local tests.

transmissions11:
the scheme for p saves gas in my tests!

Screen Shot 2022-03-21 at 11 51 56 PM

transmissions11:
want me to commit? seems like the scheme for p and the original formula for q is the fastest

recmo:
lol! In my repo it's exact opposite! Yeah, go a head an use whatever works best. But I wonder what's going on though.

src/test/FixedPointMathLib.t.sol
```diff
@@ -6,12 +6,54 @@ import {DSTestPlus} from "./utils/DSTestPlus.sol";
...
+        assertEq(FixedPointMathLib.expWadDown(-2e18), 135335283236612691);
+        assertEq(FixedPointMathLib.expWadDown(-1e18), 367879441171442321);
+
+        assertEq(FixedPointMathLib.expWadDown(-0.5e18), 606530659712633423);
+        assertEq(FixedPointMathLib.expWadDown(-0.3e18), 740818220681717866);
+
+        assertEq(
+            FixedPointMathLib.expWadDown(0),
+            999999999999999999
+            // True value: 1000000000000000000 (off by 1 wei)
```
recmo, line 22: This actually annoys me quite a bit.

src/utils/FixedPointMathLib.sol
```diff
@@ -27,6 +27,64 @@ library FixedPointMathLib {
...
+
+            // k is in the range [-61, 195].
+
+            // Evaluate using a (6, 7)-term rational approximation.
+            // p is made monic, we'll multiply by a scale factor later.
+            int256 y = x + 1346386616545796478920950773328;
+            y = ((y * x) >> 96) + 57155421227552351082224309758442;
+            int256 p = y + x - 94201549194550492254356042504812;
+            p = ((p * y) >> 96) + 28719021644029726153956944680412240;
+            p = p * x + (4385272521454847904632057985693276 << 96);
```
transmissions11, line 59: should we add a comment about using knuth's method here?

src/test/FixedPointMathLib.t.sol
transmissions11: lmk adding to the constant term works without impacting the other vars

another way thats only a couple extra opcodes would be doing 

```
           assembly {
            	r := add(r, iszero(x))
            }
```

right at the end

transmissions11:
btw how hard would it be to implement a rounding up version of this? looks like this impl truncates (which is great but i'd like to have both a floor and ceil version)

src/utils/FixedPointMathLib.sol
recmo: Sure, it's the first method from page 491. He mentions a second one which has one fewer addition, but leaves the details as an exercise. I haven't bother to do that exercise yet :)

recmo:
> btw how hard would it be to implement a rounding up version of this? looks like this impl truncates (which is great but i'd like to have both a floor and ceil version)

This should work (assuming we fix the exp(0) = 1 issue):

```solidity
function expCeil(int256 x) returns (int256) {
   if (x == 0) return 1e18; // ceil(exp(0)) = floor(exp(0)) = 1
   return exp(x) + 1;
}
```

recmo:
But I don't think it is guaranteed to return the truncated / floor result. For larger input values it doesn't have the precision required to do accurate rounding.

It also doesn't make sense to increase the precision for large inputs, because the input itself is only accurate up to 1e-18.

transmissions11:
> But I don't think it is guaranteed to return the truncated / floor result. For larger input values it doesn't have the precision required to do accurate rounding.

oh hmm yea maybe just got lucky with the test cases i used

PR #183, opened 2022-04-02
Create BeforeRedeem and AfterMint Hooks

itsdevbear:
## Description

Create Additional Internal hooks for Redeem and Mint. If you are trying to daisy chain vaults together, it is useful to have different behaviour for each hook.

Can add more tests if desired

## Checklist

- [x] Ran `dapp snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?
- [x] Ran `dapp test`?


transmissions11:
can you give me an example of why you'd need this? Not rly a fan of the complexity it adds, esp if most common use cases dont require  it

transmissions11:
oh hmm i see what you mean

itsdevbear:
Yeah so my use case is that if you want to have like a "Proxy" contract that is adhering to 4626 is able to forward its shares to a "main" vault that allows for a protocol controlled value type contract. It's for Berachain, I basically have a proxy 4626 that is deterministically created against a validators consensus address and thus I can track what validators have been delegated to.

image

itsdevbear:
@transmissions11 just following up on this king

transmissions11:
sigh this and https://github.com/Rari-Capital/solmate/pull/211 are exactly why i hate hooks, a constant battle between ppl who need more flexibility and ppl who just want simplicity (me)

transmissions11:
totally understand why u these extra hooks, im thinking maybe we impl them but just have them forward to the other hooks by default

drop19:
I am sorry if it is a bit late to jump in. But I personally find `beforeDeposit` is useful as well.

Consider the case that the token is interest bearing, and we might want to harvest interest and update `totalAssets` before the user deposit.

Each developers might need their own hooks and arguments. And I love simplicity. These hooks can be removed and devs can just override it to add hooks they want.

itsdevbear:
Shit just saw this, @transmissions11 still time to get it in for v7?

Issue #189, opened 2022-04-04
ERC1155U/ERC1155D


transmissions11:
https://etherscan.io/address/0xe76512732b1e49eb259dfc6d9d49e4c5dac6a75c#code

https://github.com/DonkeVerse/ERC1155D

PraneshASP:
Hi @transmissions11 not sure if this is the correct place for this discussion. 

Just curios to know why the `isApprovedForAll` check is ignored in the `burn()` method of  the [**Solmate's ERC1155 library**](https://github.com/Rari-Capital/solmate/blob/main/src/tokens/ERC1155.sol). Is it due to the reason that the validation can be done in the implementation contract? or is it safe to ignore the check completely?

```console
1) Contract: ERC1155Burnable
       burn
         unapproved accounts cannot burn the holder's tokens:
     AssertionError: Expected an exception but none was received
```


Because the testcases(based on OZ tests) are being reverted at the `burn` and `batchBurn` methods. Fixed the above issue by adding the validation in the implementing contract, but just wanted to check this with you as well.

Kindly advise.

Cheers!

transmissions11:
its an internal burn function, if the developer wishes to expose a public burn they must implement that behavior themselves

PR #192, opened 2022-04-07
✨ ERC1155B

transmissions11:
## Description

Closes https://github.com/Rari-Capital/solmate/issues/189

## Checklist

- [x] Ran `dapp snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?
- [x] Ran `dapp test`?


src/tokens/ERC1155B.sol
```diff
@@ -0,0 +1,272 @@
...
+        require(
+            to.code.length == 0
+                ? to != address(0)
+                : ERC1155TokenReceiver(to).onERC1155Received(msg.sender, from, id, amount, data) ==
+                    ERC1155TokenReceiver.onERC1155Received.selector,
+            "UNSAFE_RECIPIENT"
+        );
+    }
+
+    function safeBatchTransferFrom(
```
dmfxyz, line 98: 👋 I think this, `_batchMint`, and `_batchBurn` can all move the `unchecked` to the loop level

dmfxyz (commented):
✨aesthetics✨

src/tokens/ERC1155B.sol
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    function safeTransferFrom(
+        address from,
+        address to,
+        uint256 id,
+        uint256 amount,
+        bytes memory data
+    ) public virtual {
+        require(msg.sender == from || isApprovedForAll[from][msg.sender], "NOT_AUTHORIZED");
```
pcaversaccio, line 78: This comment applies to all `require` operations: why not use custom errors? I mean it decreases deployment costs for sure as well as runtime costs if the revert condition is met. Also, I think including the address of the emitter in all custom errors would make sense. This helps end users identify which contract was reverted with a failed transaction, which is especially useful for complex transactions involving multiple contracts:

```solidity
error SomeError(address emitter, ..., ..., ...);

emit SomeError(address(this), ..., ..., ...);
```

But probably you're already opinionated about custom errors 🤪

joshieDo:
Is there a rationale for using `memory` instead of `calldata` for certain function parameters (eg. arrays) ? It does seem to give better gas  results.

transmissions11:
yea memory is worse but more composable for internal use... kinda conflicted cuz anyone serious about performance will probably fork anyway

transmissions11:
well for example in `_mint`, if you make the `data` arg calldata then u can't do this

image

dmfxyz:
imo sacrificing composability in a library/inherited class is not worth the calldata savings. solmate is about gas efficiency, but not at the cost of non-optimizoors having insane headaches building off the contract.

transmissions11:
here's a compromise i think is fair: keep the internal methods (_mint) `memory` and the public ones `calldata` as those are the least likely to be called internally

PR #200, opened 2022-04-12
✅ add more precise gas tests for ERC20 transfer

alephao:
## Description

The `.gas-snapshot` file doesn't reflect the reality of how much a function run cost in gas. Take the ERC20 transfer for example:

```julia
ERC20Test:testTransfer() (gas: 60272)
```

This is pretty far from what the ERC20 transfer actually costs, so it has two downsides:

1. People wanting to compare gas costs of this implementation vs other implementations might get the wrong idea
2. When optimizing, you might get wrong comparison between the old and new gas cost because that gas is also counting other things done in the test. So for example, if we're focused on optimizing `transfer`, we'll have a lot of noise and won't get the real results since the snapshot is showing `transfer +  mint + totalSupply + balanceOf`.

This change adds a folder `benchmarks` that is supposed to have only benchmark-focused tests, or tests that only run a single function, the rest of the setup is done on `setUp` the idea is to get the gas cost as close as possible to reality. I also added an example test for ERC20 deploy and ERC20 transfer for multiple cases.

```julia
ERC20DeployBenchmarkTest:testDeploy() (gas: 787642)
ERC20TransferBenchmarkTest:testTransferWhenReceiverHasZeroTokens() (gas: 37636)
ERC20TransferBenchmarkTest:testTransferWhenReceiverOwnTokens() (gas: 20569)
ERC20TransferBenchmarkTest:testTransferWhenSenderEmptyItsPocketsAndReceiverHasTokens() (gas: 15724)
ERC20TransferBenchmarkTest:testTransferWhenSenderEmptyItsPocketsAndReceiverHasZeroTokens() (gas: 32868)
```

As you can see, the results are pretty different from the current gas snapshot.

Since this is a codebase highly focused on gas optimizations, I thought this would be a good addition. I'll be adding more of those if this change is welcome!

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?


transmissions11:
ty for your PR! imo benchmark tests will never capture the true reality of calling a function, for example in this test they'll be extra SLOADs and extcalls for the token, cheatcodes, etc. as for getting the wrong idea when optimizing, im usually not concerned with the exact magnitude of the savings, just whether gas is saved for common flows at all. ultimately for me the complexity of maintaining separate benchmark tests for such simple contracts is not worth it, so will be closing this, but an external `solmate-benchmarks` repo or similar for ppl wanting to compare our contracts with other libs would super cool to see :)

PR #212, opened 2022-04-18
⚡️ A Cheaper Sqrt Function

Gaussian-Process:
## Description

This modifies the existing sqrt implementation to remove the final three iterations from the binary search and add a linear estimate given the intermediate value. The estimate should be within a factor of ~3 of the true sqrt, which is still enough that 7 Babylonian method iterations suffice.

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?


.gas-snapshot
```diff
@@ -141,7 +141,7 @@ FixedPointMathLibTest:testMulWadDownEdgeCases() (gas: 886)
 FixedPointMathLibTest:testMulWadUp() (gas: 959)
 FixedPointMathLibTest:testMulWadUpEdgeCases() (gas: 1002)
 FixedPointMathLibTest:testRPow() (gas: 2142)
-FixedPointMathLibTest:testSqrt() (gas: 2537)
+FixedPointMathLibTest:testSqrt() (gas: 2156)
```
transmissions11, line 144: amazing

src/utils/FixedPointMathLib.sol
```diff
@@ -198,10 +198,9 @@ library FixedPointMathLib {
             // Correctness can be checked exhaustively for x < 256, so we assume y >= 256.
             // Then z*sqrt(y) is within sqrt(257)/sqrt(256) of x, or about 20bps.
 
-            // The estimate sqrt(x) = 0.75 * 0.25 * (x+1) is off by a factor of 2.67 when x=1
-            // and off by a factor of 3 when x = 256 or 1/256.
-            // In the worst case, this needs seven Babylonian iterations.
-            z := shr(12, mul(z, add(y, 65536))) // A multiply by 3 is saved from the initial z := 3
+            // The estimate sqrt(x) = (181/1024) * (x+1) is off by a factor of ~2.63 both when x=1
+            // and when x = 256 or 1/256. In the worst case, this needs seven Babylonian iterations.
+            z := shr(18, mul(z, add(y, 65536))) // A multiply is saved from the initial z := 181
```
transmissions11, line 203: is there overflow risk here?

src/utils/FixedPointMathLib.sol
```diff
@@ -165,43 +165,46 @@ library FixedPointMathLib {
...
-            // Shifting right by 1 is like dividing by 2.
+            // Now, z*z*y <= x < z*z*(y+1), and y <= 2^(16+8),
+            // and either y >= 256, or x < 256.
+            // Correctness can be checked exhaustively for x < 256, so we assume y >= 256.
+            // Then z*sqrt(y) is within sqrt(257)/sqrt(256) of sqrt(x), or about 20bps.
+
+            // The estimate sqrt(x) = (181/1024) * (x+1) is off by a factor of ~2.83 both when x=1
+            // and when x = 256 or 1/256. In the worst case, this needs seven Babylonian iterations.
+            // There is no overflow risk here since y < 2^136 after the first branch above.
+            z := shr(18, mul(z, add(y, 65536))) // A multiply is saved from the initial z := 181
```
transmissions11, line 204: forgive me for the stupid questions but where do 18 and 65536 come from? where is the `1024` thats mentioned in the comment?

src/utils/FixedPointMathLib.sol
```diff
@@ -165,43 +165,46 @@ library FixedPointMathLib {
...
-                z := shl(1, z)
-            }
-
-            // Shifting right by 1 is like dividing by 2.
+            // Now, z*z*y <= x < z*z*(y+1), and y <= 2^(16+8),
+            // and either y >= 256, or x < 256.
+            // Correctness can be checked exhaustively for x < 256, so we assume y >= 256.
+            // Then z*sqrt(y) is within sqrt(257)/sqrt(256) of sqrt(x), or about 20bps.
+
+            // The estimate sqrt(x) = (181/1024) * (x+1) is off by a factor of ~2.83 both when x=1
```
transmissions11, line 201: am i missing something or is it only off by ~0.17 at 1?

https://www.desmos.com/calculator/qxczllnaqm

transmissions11: and a lot more at 256

Gaussian-Process: so e.g. at x=1, the estimate is 0.3535, which is ~1/2.83 of the correct value, i.e. off by a multiplicative factor of 2.83
I'll make this a bit clearer in a comment

src/utils/FixedPointMathLib.sol
Gaussian-Process: I'll make this clearer in a comment too

transmissions11: awesome ty!

src/utils/FixedPointMathLib.sol
transmissions11: ahhhh gotcha my bad not familiar with that terminology

src/utils/FixedPointMathLib.sol
```diff
@@ -165,43 +165,46 @@ library FixedPointMathLib {
...
+            // and when x = 256 or 1/256. In the worst case, this needs seven Babylonian iterations.
+            // There is no overflow risk here since y < 2^136 after the first branch above.
+            z := shr(18, mul(z, add(y, 65536))) // A multiply is saved from the initial z := 181
+
+            // Run the Babylonian method seven times. This should be enough given initial estimate.
+            // Possibly with a quadratic/cubic polynomial above we could get 4-6.
```
transmissions11, line 207: whats the reason not to do this haha? esp if we could keep the iterations and remove part of the binary search (the jumps for the ifs are probably the most expensive part here)

Gaussian-Process: so I think removing even one more binary search branch is a bit tricky - that would increase the range we need to estimate sqrt on from [1/256, 256) to [1/65536, 65536). 

In that range, the best linear estimate is then off by a factor of ~12 or so, and would need 10 Babylonian iterations
If a quadratic estimate was within a factor of ~5.5 across the whole range, 8 Babylonian iterations would be enough, and maybe that's worth it?

I don't know a good way to find the best quadratic/cubic polynomials, so I'm not sure what their error bounds would look like

transmissions11: ah yeahh that bounds increase sounds like a PITA, this is already great can explore that at another time hehe

transmissions11:
amazing work thank you so much @Gaussian-Process, will get this merged asap!

src/utils/FixedPointMathLib.sol
Gaussian-Process: I tried one fewer binary branches, and the linear estimate 45/1024 with max error factor in [1/65536, 65536) of smaller than 11.4, which I think is *just* enough for 9 Babylonian iterations to suffice, but it looks like it uses slightly more gas unfortunately

transmissions11: ah interesting, thanks for investigating!

devanshbatham:
damnn, nice work @Gaussian-Process

PR #215, opened 2022-04-21
:zap: ERC1155: Use storage pointers on batch functions

joshieDo:
## Description

Use storage pointers on batch functions before loops.

## Checklist

Ensure you completed **all of the steps** below before submitting your pull request:

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?

_Pull requests with an incomplete checklist will be thrown out._

PR #217, opened 2022-04-27
♻️ Several Optimizations for Auth, ERC20, ERC1155, ERC4626

alex-ppg:
## Description

Several optimizations have been applied to the codebase to optimize the gas cost of the relevant contracts in the title. I should note that I observed another PR with part of the changes [here](https://github.com/Rari-Capital/solmate/pull/215) after I finished mine so I apologize for the duplicate but would like to further explain the optimizations.

## Checklist

- [x ] Ran `forge snapshot`?
- [ x] Ran `npm run lint`?
- [ x] Ran `forge test`?

## Rationale

I can illustrate why all optimizations are valid at the EVM level. Taking each into consideration:

### Auth: Selector Specification Optimization

The `setAuthority` function utilizes the `msg.sig` specifier to access the current call's message signature which is known at compilation time as it equals the `setAuthority(Authority)` function signature of the function itself. As there is no native EVM opcode to acquire the `msg.sig`, the instruction translates to multiple opcodes that truncate the `msg.data` payload until the `msg.sig` is derived, something that is redundant as the signature is known at compile time.

To optimize this, a minimal `IAuth` interface has been declared that contains the `setAuthority` call and thus allows it to be specified within `setAuthority` as `IAuth.setAuthority.selector`. 

### ERC20, ERC4626 & ERC1155: Mapping Lookup Optimizations

At the EVM level, all `mapping` lookup operations conduct a `keccak256` instruction underneath to calculate the proper memory offset to read the value from the `mapping` data space. As a result, a lookup such as `balanceOf[from][id]` in the case of ERC1155 would be translated to a `keccak256` instruction of `from` followed by a `keccak256` instruction of the result with `id` (Each access step performs a single autonomous `keccak256` instruction that carries the previous result over to prevent collisions).

To avoid redundant `keccak256` calculations, the result of a first-level `mapping` lookup can be cached via `storage` pointers of the relevant `mapping` type as illustrated in the ERC20, the ERC4626, and ERC1155 optimizations. 

For the ERC20 & ERC4626 cases in particular, a gas optimization is solely observed for limited approvals as an "extra" local variable is declared in cases whereby approval is unlimited.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,13 +11,12 @@ auth
 ├─ authorities
 │  ├─ RolesAuthority — "Role based Authority that supports up to 256 roles"
 │  ├─ MultiRolesAuthority — "Flexible and target agnostic role based Authority"
-mixins
-├─ ERC4626 — "Minimal ERC4626 tokenized Vault implementation"
 tokens
 ├─ WETH — "Minimalist and modern Wrapped Ether implementation"
 ├─ ERC20 — "Modern and gas efficient ERC20 + EIP-2612 implementation"
 ├─ ERC721 — "Modern, minimalist, and gas efficient ERC721 implementation"
 ├─ ERC1155 — "Minimalist and gas efficient standard ERC1155 implementation"
+├─ ERC4626 — "Minimal ERC4626 tokenized Vault implementation"
 ├─ ERC6909 — "Minimalist and gas efficient standard ERC6909 implementation"
 utils
 ├─ SSTORE2 — "Library for cheaper reads and writes to persistent storage"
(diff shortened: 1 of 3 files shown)


alex-ppg:
Hey @transmissions11, these changes should be considered as they should have a significant impact on gas costs and fit the narrative:

> Modern, opinionated, and **gas optimized**

androolloyd:
You haven't committed the gas-snapshot file that indicates the savings from these changes.

alex-ppg:
Additionally, the bytecode itself appears to be shrunk in all cases (example: `CREATE3Test:testDeployERC20`) as the code will no longer need to perform "extra" operations in the loops as well as code statements.

PR #221, opened 2022-05-02
ERC4626: add new internal virtual method withdrawAssets

0xdapper:
## Description

Add a new internal method `withdrawAssets` that might allow users to impose some withdrawal fees or any kind of fees. Thoughts?

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?


transmissions11:
thanks for the PR! but since withdrawal fees n such would need to be accounted for in other places (preview methods for example), and adding that kind of infrastructure would overcomplicate the contract i don't think its worthwhile to merge this 😓

PR #222, opened 2022-05-06
🥢trustless erc165 selector return

z0r0z:
make erc165 selector return more trustless

## Description

we can use `this.supportsInterface.selector` to be more readable and not require user verification of magick value

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?


transmissions11:
hm while this does work, prefer keeping consistency and not mixing magic values with programatic ones

PR #228, opened 2022-05-13
ERC1155 transfer trade-off

0xPhaze:
## Description

Changed balanceOf mapping from mapping(address => mapping(uint256 => uint256)) to mapping(uint256 => mapping(address => uint256)). This change reduces transfer costs, however increases mint/burn costs. Whether this trade-off is worth it is debatable, wanted to propose it nonetheless.

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?


0xPhaze:
I guess #215 makes this obsolete

PR #233, opened 2022-05-14
Add internal _transfer function to ERC721

wilsoncusack:
## Description

In this PR I add an internal `_transfer` function to the ERC721 implementation. The implementation of this is similar to [OpenZeppelin's](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/ERC721.sol#L331): It has the same functionality as transferFrom but skips the approvals checks. 

The motivation is that, in developing my own ERC721s or contracts working with them, I have found having `_transfer` very helpful. 
1. It allows for sorts of "admin override" transfers, where the contract does not care that msg.sender is approved
2. It makes overriding `transferFrom` less scary and error prone, because one can rely on calling `_transfer` for (almost) all of the core logic.   

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?


transmissions11:
60 gas extra isnt great but can live with it lol

you've worn me down on this one— will probably merge

wilsoncusack:
haha, it seems the diff is 39 in most tests? 31 in `testTransferFrom`

But epic! I would be very glad to have it. Though, make the choice you feel is best. Put this up just to have the decision documented. Would feel bad if you made the change just cause I wore you down 😀

m1guelpf:
I know DRY, but maybe an option here would be to provide the function while duplicating the implementation in transfer and transferFrom?

wilsoncusack:
@m1guelpf you might just to keep the gas diff the same? Not worth it IMO. + introduces likelihood of future mistakes/annoying to update logic in two places.

PR #234, opened 2022-05-14
Enable bi-directional conversion between bytes32 and address

kphed:
## Description

Add a method for converting a right-padded bytes32 address (e.g. return value of `fillLast12Bytes`) back into the original 20-byte address.

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?

### DIFF
--- a/src/test/Bytes32AddressLib.t.sol
+++ b/src/test/Bytes32AddressLib.t.sol
@@ -19,4 +19,13 @@ contract Bytes32AddressLibTest is DSTestPlus {
             0xCAfeBeefFeedfAceCAFeBEEffEEDfaCecafEBeeF
         );
     }
+
+    function testFromFirst20Bytes() public {
+        assertEq(
+            Bytes32AddressLib.fromFirst20Bytes(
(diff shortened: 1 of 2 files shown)


transmissions11:
awesome! can you add a fuzz test to ensure they're inverses over all inputs, and a plain unit test that just checks the output of `fromFirst20Bytes` without running it through `fillLast12Bytes`?

kphed:
> awesome! can you add a fuzz test to ensure they're inverses over all inputs, and a plain unit test that just checks the output of `fromFirst20Bytes` without running it through `fillLast12Bytes`?

Thanks for looking, done ser.

For the unit test, I'm reusing the same address-based values as what's in `testFillLast12Bytes` so I created contract variables to reduce code duplication. I wanted to bring that up since it resulted in a change to the `testFillLast12Bytes` test (now uses the new variables). If that's not ok, I'll change it back.

In the fuzz test, I have an equality assertion for the output of `fillLast12Bytes` and the bytes32-casted input because I think many people will use `fillLast12Bytes` and `fromFirst20Bytes` in tandem (we're currently doing this).

PR #184, opened 2022-04-04
⚡️ Optimize ERC721 jumps

wilsoncusack:
## Description

Rather than putting `to.code.length == 0` in a `OR` conditional in the require, we can optimize by pulling this out of the require and wrapping the require in `if (to.code.length != 0)`

## Checklist

Ensure you completed **all of the steps** below before submitting your pull request:

- [ ] Ran `dapp snapshot`?
- [ ] Ran `npm run lint`?
- [ ] Ran `forge test`?
- [ ] Ran `dapp test`?

_Pull requests with an incomplete checklist will be thrown out._
lol ok but I don't have Dapp setup on my computer

PR #237, opened 2022-05-15
⚡️ Optimize ERC-1155 JUMPs

0xpranay:
## Description

This is the 1155 version of #184. Given the zero address check for 1155, this is also more readable.

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?

yes ser


transmissions11:
good stuff, even if it didnt save gas i probably would prefer this, def more readable

Issue #242, opened 2022-05-18
Optimize merkle proof by having the arguments in calldata


hrkrshnn:
https://github.com/Rari-Capital/solmate/blob/f878f20ecd6a23dd4105ee3eed7291b6311ce1fb/src/utils/MerkleProof.sol#L9

Using `calldata` is ideal here and would save a decent chunk of gas.

https://github.com/Rari-Capital/solmate/blob/f878f20ecd6a23dd4105ee3eed7291b6311ce1fb/src/utils/MerkleProof.sol#L20-L36

But this means that the assembly routine would be more complex. Consider the following instead:

1. Use a regular for loop to load the value from `calldata`.
2. Have a subroutine called `cheapKeccak`, which uses scratch space for keccak that would do the same computation.

Vectorized:
Saves around 200 gas if there are 5 elements in proof. We can make a copy of the function just for calldata

```solidity
function merkleVerifyAssemblyCalldata(
    bytes32[] calldata proof,
    bytes32 root,
    bytes32 leaf
) internal pure returns (bool isValid) {
    assembly {
        let computedHash := leaf

        // Get the memory start location of the first element in the proof array.
        let data := proof.offset

        // Iterate over proof elements to compute root hash.
        for {
            // Left shift by 5 is equivalent to multiplying by 0x20.
            let end := add(data, shl(5, proof.length))
        } lt(data, end) {
            data := add(data, 0x20)
        } {
            let loadedData := calldataload(data)
            // Slot of `computedHash` in scratch space.
            // If the condition is true: 0x20, otherwise: 0x00.
            let scratch := shl(5, gt(computedHash, loadedData))
            
            // Store elements to hash contiguously in scratch space.
            // Scratch space is 64 bytes (0x00 - 0x3f) and both elements are 32 bytes.
            mstore(scratch, computedHash)
            mstore(xor(scratch, 0x20), loadedData)
            computedHash := keccak256(0x00, 0x40)
        }
        isValid := eq(computedHash, root)
    }
}
```

transmissions11:
tryna think of a case where u need the proof to be `memory`? im totally cool with just supporting calldata

hrkrshnn:
@Vectorized Do you really have to use `calldataload(..)` in assembly? Do you have a benchmark for just `proof[idx]`?

Vectorized:
@hrkrshnn I tried `proof[idx]`, but the assembly cannot compile.

Here are some numbers on Forge:
```
[...]

PR #243, opened 2022-05-18
⚡️ MerkleProof Verify Calldata Version

Vectorized:
## Description

This is for #242. 

After some marinating, I've decided to change verify to take in a calldata array instead of a memory array.

Since MerkleProof is used for saving on-chain storage afterall,  
the data is almost certainly likely to be from an external call.

Since Solmate is an opinionated minimalist repo, I'd recommend just including the version that is most likely to be relevant.

Also, sorry for missing the empty trailing space in the previous PR.

## Checklist

- [X] Ran `forge snapshot`?
- [X] Ran `npm run lint`?
- [X] Ran `forge test`?

## Gas Benchmarks

The gas snapshots will have higher gas because of the call overhead required to test calldata.  

Once you factor in the call overhead, these are the numbers:

```
[PASS] testValidProofSupplied() (gas: 2464)
[PASS] testValidProofSuppliedCalldata() (gas: 2280)
[PASS] testVerifyEmptyMerkleProofSuppliedLeafAndRootDifferent() (gas: 1733)
[PASS] testVerifyEmptyMerkleProofSuppliedLeafAndRootDifferentCalldata() (gas: 1565)
[PASS] testVerifyEmptyMerkleProofSuppliedLeafAndRootSame() (gas: 1727)
[PASS] testVerifyEmptyMerkleProofSuppliedLeafAndRootSameCalldata() (gas: 1537)
[PASS] testVerifyInvalidProofSupplied() (gas: 2460)
[PASS] testVerifyInvalidProofSuppliedCalldata() (gas: 2322)
```


src/utils/MerkleProof.sol
```diff
@@ -37,4 +37,27 @@ library MerkleProof {
...
+            } {
+                let loadedData := calldataload(data)
+                let scratch := shl(5, gt(computedHash, loadedData))
+                mstore(scratch, computedHash)
+                mstore(xor(scratch, 0x20), loadedData)
+                computedHash := keccak256(0x00, 0x40)
+            }
+            isValid := eq(computedHash, root)
+        }
+    }
```
transmissions11, line 62: mind adding some comments here?

src/utils/MerkleProof.sol
```diff
@@ -37,4 +37,27 @@ library MerkleProof {
             isValid := eq(computedHash, root)
         }
     }
-}
\ No newline at end of file
+
+    function verifyCalldata(
```
transmissions11, line 41: imo lets rename this to verify and the other to verifyMemory

src/test/MerkleProof.t.sol
```diff
@@ -37,8 +37,56 @@ contract MerkleProofTest is DSTestPlus {
...
+        return MerkleProof.verify(proof, root, leaf);
+    }
+
+    function verifyCalldata(
+        bytes32[] calldata proof,
+        bytes32 root,
+        bytes32 leaf
+    ) external pure returns (bool) {
+        return MerkleProof.verifyCalldata(proof, root, leaf);
+    }
```
transmissions11, line 91: these feel kinda pointless unless we make any assertions?

src/utils/MerkleProof.sol
transmissions11: or actually can we just remove verifyMemory?

Vectorized: Let's remove the memory version.

Vectorized:
Ok, remove the memory version.   
This will also indirectly enforce good practice.

Since Solmate is an opinionated minimalist repo, it makes more sense to have just one recommended version.

If someone somehow needs the memory version, they are always free to copypasta it and modify.

transmissions11:
amazing work, lgtm!

PR #254, opened 2022-05-25
⚡ MerkleProof Minor Optimizations

Vectorized:
## Description

We got to optimize a little bit more.

**Side note:** 

Verifying the Merkle proof is a task where Solidity does not have a clear lead over Vyper.

The current code example has a lead of a few hundred gas over a Vyper implementation when the tree's height (i.e. `proof.length`) is 10. But for shorter heights like 2, the Vyper implementation has a lead of a few hundred gas.

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?


Vectorized:
Let this marinate a while first. I need to double check the multi merkle proof. 

To on-lookers, don't use yet.

If need be, we canb just scrape the multi proof part.

I think, the source JS and TS codes have some unhandled edge cases.

```javascript
it.only('test valid empty Merkle multi proof', async function () {
  const leaves = ['a'].map(keccak256).sort(Buffer.compare);
  const merkleTree = new MerkleTree(leaves, keccak256, { sort: true });

  const root = merkleTree.getRoot();
  const proofLeaves = ['a'].map(keccak256).sort(Buffer.compare);
  const proof = merkleTree.getMultiProof(proofLeaves);
  const proofFlags = merkleTree.getProofFlags(proofLeaves, proof);

  console.log({
    description: 'returns true for a valid Merkle multi proof',
    root: root, 
    leafs: proofLeaves, 
    proofs: proof,
    proofFlags: proofFlags
  });
  
  // Throws exception, when it should return true.
  console.log(merkleTree.verifyMultiProof(root, proofLeaves, proof, proofFlags)); 
  // Throws exception, when it should return true.
  expect(await this.merkleProof.multiProofVerify(root, proofLeaves, [], [])).to.equal(true);
});
```

@miguelmota

joshieDo:
Bad CI!
```
Failed tests:
[FAIL. Counterexample: calldata=0x94a508f10000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000c000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000000, args=[true, true, false, true, true, [false]]] testMultiProofVerifyEmpty(bool,bool,bool,bool,bool,bool[]) (runs: 1513, μ: 36734, ~: 37585)
```

Vectorized:
@joshieDo Thanks. Fixed the fuzz test.

Vectorized:
https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3446

PR #264, opened 2022-06-05
Bytes32AddressLib Save Gas

itsdevbear:
## Description
Save some gas by utilizing the free memory pointer space (which will be zero'd out) to just directly load the calldata and then return it as is.

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?


transmissions11:
this is an internal function, the inputs are not guaranteed to be in calldata at all, let alone at those specific offsets?

transmissions11:
also using the return statement in yul does not return from the local callframe, it ends the top level call entirely (like reverting)

itsdevbear:
Oh 100% facts, was too on autopilot

transmissions11:
hehe

PR #278, opened 2022-06-23
⚔️ functions clashing with forge-std/Test.sol

vminkov:
## Description

These functions are now also implemented almost identically in `forge-std/Test.sol` and they make it impossible to inherit both `DSTestPlus` and `Test` , like it is done in the `fuse-flywheel` repo on a few occasions:
https://github.com/fei-protocol/fuse-flywheel/blob/b1ebb618fe2965727d5a81d74d374a667a1e52b9/src/test/integration/FlywheelDynamicRewardsIntegration.t.sol#L25
https://github.com/fei-protocol/fuse-flywheel/blob/b1ebb618fe2965727d5a81d74d374a667a1e52b9/src/test/integration/pools/FusePool156test.t.sol#L16
https://github.com/fei-protocol/fuse-flywheel/blob/b1ebb618fe2965727d5a81d74d374a667a1e52b9/src/test/integration/pools/FusePool190test.t.sol#L15

*This is a draft PR just to gather opinions and the checklist below has not been done*

## Checklist

- [ ] Ran `forge snapshot`?
- [ ] Ran `npm run lint`?
- [ ] Ran `forge test`?

### DIFF
--- a/src/test/utils/DSTestPlus.sol
+++ b/src/test/utils/DSTestPlus.sol
@@ -63,12 +63,7 @@ contract DSTestPlus is DSTest {
         emit log_named_uint(string(abi.encodePacked(checkpointLabel, " Gas")), gasDelta);
     }
 
-    function fail(string memory err) internal virtual {
-        emit log_named_string("Error", err);
-        fail();
-    }
-
-    function assertFalse(bool data) internal virtual {
+    function _assertFalse(bool data) private {
         assertTrue(!data);
     }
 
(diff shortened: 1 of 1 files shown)


zerosnacks:
My personal opinion would be that we deprecate `DSTestPlus` in favor of `forge-std/Test.sol` (https://github.com/foundry-rs/forge-std) as it supersedes `DSTestPlus`. If we do want to ship some extensions that we cannot get merged upstream we could still extend it here using a `Fixture`.

transmissions11:
they're not meant to be used together right now, that will change in our V7 release but im not gonna make sweeping changes like this to master rn

PR #280, opened 2022-06-24
fix(solidity): avoid && logic in req. statements

sambacha:
Avoid && logic in require statements

Can gas golf to show benefit if needed

### DIFF
--- a/src/tokens/ERC20.sol
+++ b/src/tokens/ERC20.sol
@@ -151,7 +151,8 @@ abstract contract ERC20 {
                 s
             );
 
-            require(recoveredAddress != address(0) && recoveredAddress == owner, "INVALID_SIGNER");
+            require(recoveredAddress != address(0), "INVALID_SIGNER");
+            require(recoveredAddress == owner, "INVALID_SIGNER");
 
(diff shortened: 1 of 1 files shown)


transmissions11:
yeah i know about this optimization (why solc???? why...) but i dont think the gas savings are worth the readability savings

sambacha:
> thanks for the PR!
> 
> 
> 
> i've seen this optimization before (why solc???? why...) but i dont think the gas savings are worth the readability cost

Wait, you think this is less readable?

sambacha:
> thanks for the PR!
> 
> 
> 
> i've seen this optimization before (why solc???? why...) but i dont think the gas savings are worth the readability cost

Either way thanks for the response, cheers

z0r0z:
can't we avoid this double check regardless by only doing `require(recoveredAddress == owner)`?

is `!= address(0)` doing any work? For example, as I understand, address(0) is recovered if things fail on invalid signature. This gets scooped into == owner check? And if UI messes up and puts address(0) in as the owner, there would be no valid signature, and the same thing will happen anyways?

PR #282, opened 2022-06-30
✂️Snip gas tax in ERC20 for null owner permit

z0r0z:
## Description

remove gas tax for 'risk' of UI inputting owner == address(0). While technically this doesn't match eip2612 call for revert in cases where owner == address(0), kinda feels silly and can make that edit to eip right

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?

### DIFF
--- a/src/tokens/ERC20.sol
+++ b/src/tokens/ERC20.sol
@@ -151,7 +151,7 @@ abstract contract ERC20 {
                 s
             );
 
-            require(recoveredAddress != address(0) && recoveredAddress == owner, "INVALID_SIGNER");
+            require(recoveredAddress == owner, "INVALID_SIGNER");
 
             allowance[recoveredAddress][spender] = value;
         }


0xdapper:
I think this check still needs to be there because without this people can recover potentially burned tokens? `ecrecover` precompile would return `0x0` if recovery fails, so i could just give garbage hash, v, r, s and permit myself to claim all the recovered tokens.

jaglinux:
@z0r0z , can you upload the gas snapshot, any tests covers this case so that we can see the gas reductions.

transmissions11:
although address(0) isn't owned by anyone so being able to steal from it *feels* ok, not being able to steal tokens like that is a fundamental erc20 invariant and brekaing could lead to some confusing situations that i'd rather avoid

will have pass on this one, but i appreciate the creativity @z0r0z <3

PR #283, opened 2022-07-02
feat: Move ERC4626 Mixin to Tokens

refcell:
## Overview

Remove `mixins` and `ERC4626.sol` to [ERC4626](./src/tokens/ERC4626.sol).


devtooligan:
i like this change.  its true that its a "mixin" since it imports erc20 but i dont think thats an important distinction to make a separate folder for.  i think normally people think of 4626 as a token not as a mixin so having it in that folder probably hides its existence from some people. 

changing the folder like this will encourage higher usage of this contract as well as greater adoption of 4626.

also dreezy forgot to attach a cute animal pic so:
dogggy

abigger87:
fixed with the utmost haste
qt

PR #320, opened 2022-09-18
Add compact string encoding library

d1ll0n:
## Description

Adds utils/PackedStringLib.sol - a library for efficiently encoding strings that are <=31 bytes.
- Encodes input strings by putting the length in the first byte and the body in the following 31 bytes.
- Decodes strings by clearing two words in memory (to ensure the string has leading and trailing zeroes) and then writing the packed form to the last byte of the length buffer (first word), which also places the right-padded body of the string where it needs to be.

[Solidity's built-in mechanism](https://docs.soliditylang.org/en/v0.8.17/types.html#bytes-and-string-as-arrays) for accessing storage strings is efficient when strings can be any size, but in practice most tend to be quite small and static. This means that all of the additional code for determining the type of string (single or multiple word), looping over words and hashing each index to get the slot, goes totally unused. Surprisingly, this can add up to several hundred (or even thousand!) bytes depending on your optimization settings, and can really eat into available contract space.

Below is a comparison of the simplest use-case (retrieving a contract's name) with packed vs. standard strings and constant/immutable vs. storage.

contracts

I tried to stick to the coding conventions as much as possible, but this file has a lot more comments than most - had a bit of an ADHD moment. @transmissions11 if you prefer to keep to the clean aesthetic I'm happy to remove them.

I'm also a little new to Forge, so I'd appreciate a second opinion on whether there are any gaps in the testing.

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?


src/utils/PackedStringLib.sol
```diff
@@ -0,0 +1,126 @@
...
+/// code that would have never been executed.
+/// @notice https://docs.soliditylang.org/en/v0.8.17/types.html#bytes-and-string-as-arrays
+/// @author Solmate (https://github.com/transmissions11/solmate/blob/main/src/utils/PackedStringLib.sol)
+library PackedStringLib {
+    error UnpackableString();
+
+    /// @dev Pack a 0-31 byte string into a bytes32.
+    /// @dev Will revert if string exceeds 31 bytes.
+    function packString(string memory unpackedString) internal pure returns (bytes32 packedString) {
+        uint256 length = bytes(unpackedString).length;
```
hrkrshnn, line 20: Can't you do something like `bytes32(bytes(unpackedString))`?

The conversion would truncate the string. So you'd probably need the `if (length > 31)` check if truncation is not good enough.

hrkrshnn: Also would be nice to define a custom type `type ShortString is bytes32;`

PR #330, opened 2022-10-06
⚡️ erc1155 fn args memory -> calldata

greg-schrammel:
## Description

set function arguments as `calldata` since they are not modified, removing the need to cache some array lengths

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?


transmissions11:
this prevents passing arrays generated in memory to these functions

PR #333, opened 2022-10-14
⚡️ Memory safe assembly

axic:
## Description

Mark assembly blocks `memory-safe`, so that they can be used with optimizer stages (for example the stack-to-memory elevation). See https://docs.soliditylang.org/en/v0.8.17/assembly.html?highlight=memory-safe#memory-safety for details.

I did spend quite some time manually reviewing each case in this codebase, but I don't feel confident merging this without at least 2 proper reviews (cc @hrkrshnn).

## Checklist

- [ ] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [ ] Ran `forge test`?

### DIFF
--- a/src/utils/SSTORE2.sol
+++ b/src/utils/SSTORE2.sol
@@ -34,7 +34,7 @@ library SSTORE2 {
             runtimeCode // The bytecode we want the contract to have after deployment. Capped at 1 byte less than the code size limit.
         );
 
-        assembly {
+        assembly ("memory-safe") {
             // Deploy a new contract with the generated creation code.
             // We start 32 bytes into the code to avoid copying the byte length.
             pointer := create(0, add(creationCode, 32), mload(creationCode))
@@ -79,7 +79,7 @@ library SSTORE2 {
         uint256 start,
         uint256 size
     ) private view returns (bytes memory data) {
-        assembly {
+        assembly ("memory-safe") {
             // Get a pointer to some free memory.
             data := mload(0x40)
 
--- a/src/utils/SSTORE2.sol
+++ b/src/utils/SSTORE2.sol
@@ -34,7 +34,8 @@ library SSTORE2 {
             runtimeCode // The bytecode we want the contract to have after deployment. Capped at 1 byte less than the code size limit.
         );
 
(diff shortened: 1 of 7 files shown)


src/utils/SSTORE2.sol
```diff
@@ -79,7 +79,7 @@ library SSTORE2 {
         uint256 start,
         uint256 size
     ) private view returns (bytes memory data) {
-        assembly {
+        assembly ("memory-safe") {
```
hrkrshnn, line 82: There can be dirty values at the end of the byte string btw.

Need to clean the last memory chunk here.

Take size to be 31 and have all the memory to be dirty. The last bit of the `data` would be dirty.

axic: Yes, that's correct, but shouldn't affect `memory-safe`?

hrkrshnn: Yeah. This is independently of memory safe. I'll separate it into a new issue.

axic:
@transmissions11 ready for review, will squash once we get closer to merging.

transmissions11:
isn't this `(memory-safe)` modifier for the assembly block only available in more recent versions of 0.8? since we want to maintain compatibility with all versions we should probably use the comment syntax, no?

axic:
> isn't this (memory-safe) modifier for the assembly block only available in more recent versions of 0.8? since we want to maintain compatibility with all versions we should probably use the comment syntax, no?

Good point, can change to that. Hmm, actually wonder how compatible that is given it starts with `@solidity` as a new natspec tag 😬 

Btw, the `assembly <string literal> {}` syntax was valid since assembly was introduced, but it only accepted `evmasm` as the sole value plus the parenthesis syntax is new.

axic:
@transmissions11 what is the earliest 0.8.x version do you want to support? There should be a CI run for that version, just to be sure.

transmissions11:
> @transmissions11 what is the earliest 0.8.x version do you want to support? There should be a CI run for that version, just to be sure.

This is a good point, ideally 0.8.0 itself. I'll open an issue.

PR #335, opened 2022-10-25
⚡Optimize gas use of safe ERC20 methods in SafeTransferLib

Philogy:
## Description

Optimize the `returndatasize >= 31` check in the `SafeTransferLib` library's `safeTransferFrom`, `safeTransfer` and `safeApprove` methods. This is done by reusing the `eq(mload(0), 1)` check. If memory location 0x00 is zeroed-out before the call the only way it can be `1` after the call is if the length of the returndata was at least 32. Saves 2 gas. 

An alternative implementation of this optimization would be to use the zero memory slot 0x60 for the returndata, this would allow moving the `mstore(0x60, 0)` after the require, reducing gas cost in the revert case while having the same cost in the success case. The downside to this approach would be the slightly higher bytecode size of the contract.

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?


Philogy:
Closed due to the overlooked #259 which has a better version of this optimization

PR #340, opened 2022-11-04
✅ Fix test to fail from old deadline not arithmetic underflow

heyskylark:
## Description
In `ERC20.t.sol` it appears the test `testFailPermitPastDeadline` is failing due to an underflow error from subtracting `block.timestamp` of 0 by 1. The test fails successfully, but not from the expired deadline of the permit.

To fix the issue, I set the timestamp in the signature and the permit input to the original block.timestamp of 0. Then I used `hevm.warp` to increment the current `block.timestamp` by 1.

## Checklist

- [X] Ran `forge snapshot`?
- [X] Ran `npm run lint`?
- [X] Ran `forge test`?


transmissions11:
ah nice great find.

transmissions11:
can you run `forge snapshot`?

PR #369, opened 2023-04-11
move loop invariant variables out from loop

MollyHe5:
## Description

Moving loop invariant codes out from loop can save gas.
In the following demo, it can save 765 units of gas when the length of ```arr``` is 100.

```
    function test1(uint256[] calldata arr) public {     
        for(uint256 i = 0; i < arr.length; i++) {
        }
    }

    function test2(uint256[] calldata arr) public {
        uint256 len = arr.length;
        for(uint256 i = 0; i < len; i++) {
        }
    }
```

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?


transmissions11:
these are calldata variables, so caching them on the stack is actually more expensive in practice.

songlh:
@transmissions11 can you take a look at @MollyHe5 's response?

transmissions11:
a) the length of the array is a calldata variable

b) any positive differences you're seeing in gas there are just artifacts of testing via two separate functions which have different precedence in the function dispatcher  

c) the gas snapshot diff associated with this PR is evidence of the above, gas goes up slightly for `testBatchBalanceOf` and `testSafeBatchTransferFromToEOA`, etc

PR #381, opened 2023-08-08
⚡️ Optimize `SignedWadMath` edge case check

Philogy:
## Description

I've changed the `SignedWadMul.wadMul` to do the overflow & edge case check in one `if`:

```yul
if iszero(
    and(
        or(iszero(x), eq(sdiv(r, x), y)),
        or(lt(x, not(0)), sgt(y, 0x8000000000000000000000000000000000000000000000000000000000000000))
    )
) {
    revert(0, 0)
}
```

This saves gas at runtime for all cases except the edge case because the resulting bytecode only uses 1 `JUMPI` vs. 2 to check the conditions. Branches are currently quite expensive making it worth the slightly higher cost of combining multiple expressions.

To save further gas I've transformed the `!(x == -1 && y == type(int256).min)` check into `x != -1 || y != type(int256).min` to minimize the use of `ISZERO`. The inequality checks both need to have binary results so the resulting bitwise expression `and(no_overflow, or(x_not_neg1, y_not_min_signed))` equates the logical expression across the whole set of values. To achieve this `LT` and `SGT` are abused as `!=` operators, this is possible here as `-1` represents the highest possible unsigned integer and `0x8000000000000000000000000000000000000000000000000000000000000000` the lowest signed number.

The final inline-assembly expression represents:

```solidity
require(
    (x == 0 || (r / x == y))
    && (x != -1 || y != type(int256).min)
);
```

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?


transmissions11:
nice, lgtm!

PR #387, opened 2023-09-03
ci(forge): fix failing workflow

sambacha:
The previous commit saw a CI failure due to not having `--tolerance` specified. I took the difference and rounded up (~2.64% difference in gas snapshot to the value of 3). 

Also updated the foundry tool chain to the appropriate repo, and updated checkout to use v3.

## Description

Describe the changes made in your pull request here.

## Checklist

- [ ] Ran `forge snapshot`?
- [ ] Ran `npm run lint`?
- [ ] Ran `forge test`?

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -6,10 +6,10 @@ jobs:
   tests:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       - name: Install Foundry
-        uses: onbjerg/foundry-toolchain@v1
+        uses: foundry-rs/foundry-toolchain@v1
         with:
           version: nightly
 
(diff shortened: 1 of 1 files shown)


transmissions11:
thanks! but what happened to having a deterministic seed?

sambacha:
> thanks! but what happened to having a deterministic seed?

You want to create a secondary check with tolerance if the original check fails then?

transmissions11:
hm? forge used to use a deterministic random seed such that re-running would always give the same result and we wouldn't need tolerance at all

sambacha:
> hm? forge used to use a deterministic random seed such that re-running would always give the same result and we wouldn't need tolerance at all

[https://book.getfoundry.sh/reference/config/testing?highlight=seed#seed](https://book.getfoundry.sh/reference/config/testing?highlight=seed#seed)

> *   Type: string (hexadecimal)
> *   Default: none
> *   Environment: `FOUNDRY_FUZZ_SEED`
> 
> Optional seed for the fuzzing RNG algorithm.

I think this changed with the `revm` switchover, not sure tbh.

PR #390, opened 2023-09-30
Update ERC1155.sol:⚡Optimized For loops

Aryan9592:
It be great not to initialize `uint256 i` to 0, cuz the default values of all unsigned integers usually stick to 0..just a good practice I guess.

Refer this -> https://gist.github.com/grGred/9bab8b9bad0cd42fc23d4e31e7347144#for-loops-improvement

**These changes won't really bring out some bugs or errors. Thus, I didn't felt any need of running the tests which were mentioned!**

### DIFF
--- a/src/tokens/ERC1155.sol
+++ b/src/tokens/ERC1155.sol
@@ -90,7 +90,7 @@ abstract contract ERC1155 {
         uint256 id;
         uint256 amount;
 
-        for (uint256 i = 0; i < ids.length; ) {
+        for (uint256 i; i < ids.length; ) {
             id = ids[i];
             amount = amounts[i];
 
@@ -128,7 +128,7 @@ abstract contract ERC1155 {
         // Unchecked because the only math done is incrementing
         // the array index counter which cannot possibly overflow.
         unchecked {
-            for (uint256 i = 0; i < owners.length; ++i) {
+            for (uint256 i; i < owners.length; ++i) {
                 balances[i] = balanceOf[owners[i]][ids[i]];
             }
         }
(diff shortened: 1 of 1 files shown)


transmissions11:
my understanding is this does not save gas, and is slightly unintuitive, so does not feel worth it to me

Aryan9592:
As your wish!!

Aryan9592:
Just a query for learning point of view:

Here are two code snippets from this solidity file:

```
// Unchecked because the only math done is incrementing
// the array index counter which cannot possibly overflow.
unchecked {
    for (uint256 i = 0; i < owners.length; ++i) {
        balances[i] = balanceOf[owners[i]][ids[i]];
    }
}
```

and

```
for (uint256 i = 0; i < ids.length; ) {
    id = ids[i];
    amount = amounts[i];
    
     balanceOf[from][id] -= amount;
     balanceOf[to][id] += amount;

     // An array can't have a total length
     // larger than the max uint256 value.
     unchecked {
         ++i;
     }
}
```

In the first code snippet, you dropped down the whole `for` loops inside that `unchecked` block, but in the second you just kept that incrementing factor `++i`
Like why it is done in this way, why apply the same method in both...Well, I know it's due to some overflow errors, but still want to know it by depth

@transmissions11 feel free to answer this, if you got time!

malik672:
@Aryan9592 because two can probably have underflows and overflows why one can't, i mean look at both loops, one of them performs an arithmetic operation in it, why the other does not

Aryan9592:
got it!

PR #405, opened 2023-12-14
Change value for short path to return zero in `wadExp`

engn33r:
## Description

This is the same optimization described in https://github.com/pcaversaccio/snekmate/pull/189

This PR is for a very small optimization that is possible by increasing the range of values that will take the short logic path of `wadExp()` and return zero. I am not sure how the previously chosen value of -42139678854452767551 was selected, but these tests demonstrate the actual turning point when the result of the current `wadExp()` implementation transitions from 1 to zero.

```solidity
  function testWadExpZeroPoint() public {
      assertEq(wadExp(-41446531673892822312), 1);
      assertEq(wadExp(-41446531673892822313), 0);
  }
```

The change to CREATE3.t.sol is from the linter, not a change I deliberately made.

## Checklist

- [X] Ran `forge snapshot`?
- [X] Ran `npm run lint`?
- [X] Ran `forge test`?


transmissions11:
interesting, nice find!

cc @recmo — want to double check, does this look right to you?

engn33r:
For reference, you can get this value in python (minus some precision) with `numpy.log(10**-18)`

